PR #6, opened 2016-12-23
Bicop refactoring

tnagler:
So I did some major refactoring of the `Bicop` stuff. It is compatible with the master branch, but **may not be compatible with what you have done in the meantime**. So don't merge this request right now, we can discuss and change a few things.

I'll try to summarize the changes:

1. Class `Bicop` now has private members `family_`, `rotation_`, `parameters_`,  `parameter_bounds_`, getters for all of them, and setters for all but `family_` ([bicop.hpp](https://github.com/tvatter/vinecopulib/blob/bicop-refactoring/src/common/include/bicop.hpp)). Removed corresponding code where redundant.

1. Added default `parameter_bounds` to all family constructors.

1. Added some constructors, e.g., `IndepBicop(const VecXd& parameters)`,  `IndepBicop(const VecXd& parameters, const int& family)`. They could become useful when we want to initialize a `Bicop` object without knowing their family. They could also be useless ;)

1. Added function `calculate_tau()` to `Bicop` class. It calculates the Kendall's tau for the current `family_` and `parameters_` specification by wrapping around `par_to_tau()`

1. Make `tau_to_par()` always return `VecXd`. It would get ugly if we need to check if its `double` or `VecXd`.

1. Added two convenience header files:

     -  [bicop_families.hpp](https://github.com/tvatter/vinecopulib/blob/bicop-refactoring/src/common/include/bicop_families.hpp): includes all family headers.

    -  [bicop.hpp](https://github.com/tvatter/vinecopulib/blob/bicop-refactoring/src/common/include/bicop.hpp): includes the class and familie headers.

1. Use same naming convention for alll include guards: a file `tik_tok.hpp` gets `VINECOPLIB_TIK_TOK_HPP`. 

1. Renamed `NormalBicop` to `GaussBicop` (I think this name is used more commonly).

1. Adjusted all test files accordingly.

1. Some changes regarding coding style and readability.


tvatter:
My comments (not by order of importance):
- For `calculate_tau()`, which is now in `bicop_class.hpp`, can you guarantee that this will be also implemented for non-parametric families? If yes, then it's a very good idea!
- Concerning the naming convention `VINECOPLIB_TIK_TOK_HPP` should be `VINECOPULIB_TIK_TOK_HPP`. Similarly, the license preambles in everyfile still mention "vinecoplib" and should be updated to "vinecopulib". Also, add Thomas Nagler to the copyright.
- In `bicop.hpp`, I think that the inclusion of `bicop_class.hpp` is useless, since including `bicop_families.hpp` include the headers of the families (and each of them already includes `bicop_class.hpp`). As such, I also think that the `bicop.hpp` header itself is useless, since the same information is already included in `bicop_families.hpp`.
- In `bicop_parametric.hpp`, the function `calculate_tau()` hides a non-virtual function (from `bicop_class.hpp`) but is not implemented. I think that this declaration should be removed (again, provided that the `calculate_tau()` function can be implemented at the Bicop if you can guarantee that it's OK for non-parametric families).
- Concerning `calculate_npars`, why isn't it implemented at the Bicop level (since you moved the `parameters_` data member at the Bicop level)? Is it because, for non-parametric families, the number of parameters won't be equal to the size of this parameters vector?
- Concerning moving the `parameters_` data member at the Bicop level, are you sure that it is a good idea for non-parametric families? I thought that such families where represented by a grid of points to interpolate over (i.e., two matrices).
- Please remove `get_bounds_standard` from `bicop_claton.hpp`.
- I am still pondering about the choice of handling the rotation at the Bicop level. Does it make sense for non-parametric families? Does it make sense for elliptical copula? In any case, at the same level where the rotation is handled, we should probably also be careful with the pdf, hfuncs and hinvs, in the same manner as it is now implemented in `bicop_archimedean.cpp`.
- If the parameters of archimedean copulas, then the pdf/hfunc/hinv in `bicop_archimedean.cpp` needs to be modified (e.g. at line 43).

tnagler:
Answers (same order as your comments, will give them numbers so we can reference them):

1. Since the Kendall's tau can be expressed in terms of the copula density (and double integral), we can write a generic fall-back that works whenever no analytic solution is available.

1. You're obviously right about `VINECOPULIB_TIK_TOK_HPP` and the license preambles (I just replaced the first instance). Will add myself to the copyright.

1. You're right about `bicop.hpp`. Should I just rename `bicop_families.hpp` to `bicop.hpp`. I think it's a good idea to have a header that imports all `Bicop`-related functionality.

1. Will remove `calculate_tau()` in `bicop_parametric.hpp`.

1. That's the reason exactly, for the 'number of parameters'  of the nonparametric families, we will need a family specific implementaiton.

1. Generally, nonparametric estimators are characterized by a vector of smoothing parameters and the data. The `kdecopula` package uses spline interpolation as a workaround to storing all the data. Bernstein and B-spline copulas have the number of knots as the smoothing parameter and one can store the basis coefficients instead of the data. In any case the `parameters_` should refer to the smoothing parameter and anything else should be additional data members at the family level.

1. Will do.

1. I'm not sure if it is necessary for elliptical and nonparametric families. But it's certainly well defined. I want to have it at the `Bicop` because that gives us a unified API for creating and manipulating `Bicop` instances, irrespective of their family. At compile time, we may not know the family, but should still be able to write code that works for all families.  
You're certainly right about handling the rotation for `pdf()`, `hfuncx()` and `hinvx()` similar to `bicop_archimedean.cpp`.

1. I guess you mean that all parameters are positive (irrespective of rotation). Then yes, this has to be modified.

1. You're right, that's more consistent. Will change that. 

1. That's a leftover from finding bugs, will remove the comments.

tvatter:
11. OK
13. Both options are fine with me. I agree that this is a good idea.
15. OK
16. OK
17. I don't mind having them here as long as the rotations are handled properly.
18. Yes, that's what I meant.

tvatter:
Another thing: in the constructor by default of the student bicop, can you add the following?

    parameter_bounds_ = MatXd::Ones(2, 2);
    parameter_bounds_(0, 0) = -1.0;
    parameter_bounds_(1, 0) = 2.0;
    parameter_bounds_(1, 1) = 50.0;

tvatter:
One last thing:

In `clayton_bicop.cpp` in `tau_to_parameters`

PR #9, opened 2016-12-27
Rotation handling

tnagler:
The rotations are now handled at the `Bicop`-level. The implementation at the `Bicop`-level relies on  `pdf_default`, `hfunc1_default`, and so on which constitute the family-specific implementations of the pdf and (inverse) h-functions for the case `rotation_ = 0`.

I'm not too happy with the `_default` suffix, suggestions are welcome ;)

### DIFF
--- a/src/common/bicop_archimedean.cpp
+++ b/src/common/bicop_archimedean.cpp
@@ -18,8 +18,6 @@ along with vinecopulib.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 #include "include/bicop_archimedean.hpp"
-#include <iostream>
-
 
 // PDF
 VecXd ArchimedeanBicop::pdf_default(const MatXd& u)
--- a/src/common/bicop_archimedean.cpp
+++ b/src/common/bicop_archimedean.cpp
@@ -18,152 +18,45 @@ along with vinecopulib.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 #include "include/bicop_archimedean.hpp"
+#include <iostream>
 
 
 // PDF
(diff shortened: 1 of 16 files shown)


tvatter:
- I think that the `XXX_default` methods should be protected and that only the actual methods should be public. Furthermore, since, for both archimedean and elliptical copulas, `hXXX2_default` are implemented using `hXXX1_default(swap_cols(u))`, can’t `hXXX1/2` be written directly using only `hXXX1_default`?  If that is the case, couldn’t we have only `hXXX_default` as protected methods defined at the concrete class level and `hfunc1/2` and `hinv1/2` at the Bicop level?
- Why did you add `#include <iostream> in bicop_archimedean.hpp`?
- I do not understand what the `needs_check_` is for in the unit tests. It is set to `true` by default and to `false` if we deal with family in {3,4,6} and the `rotation_` data member in the Instance object is not equal to zero, correct? Furthermore, when `needs_check_ = false`, then nothing is done, correct? If yes, why is that? If no, what am I missing?

tnagler:
- Portecting the default methods is a good idea. It's also true that it's superfluous to have `hXXX1_default` and `hXXX2_default` for the elliptical and Archimedean classes. But it will be necessary for more asymmetric families (especially nonparametric ones). To ensure that rotations can be handled generically at the Bicop level for all families, we need to define both `hXXX1_default` and `hXXX2_default`.
- That's a left-over from debugging.
- The variable `is_rotless` is true when the family is in {0, 1, 2, 5}. Consequently, `needs_check_ = false` if and only if family is in {0, 1, 2, 5} and `rotation_` is not zero.

tvatter:
- OK
- OK
- OK

Looks good then, I'm merging!

PR #10, opened 2016-12-27
Bicop factory

tnagler:
_Warning: This PR also contains all changes suggested in #9 which should be merged first._

Here's my implementation of a factory for the `Bicop` class. It returns a `std::shared_ptr<Bicop>` (typedef'd as `Bicop_ptr`).


tvatter:
Well, for this one, I must admit that the factory implemented is not at all what I was thinking about. In my opinion, the `create` method should rather be called `select` and "mimic" the behaviour of `BiCopSelect` from the `Vinecopula` package. As such, I was thinking about a factory whose arguments would be `data`, `familyset`, `selectioncrit`, `rotations`, `presel` and `method`. I will implement such a factory and make a commit in this branch, and we can further discuss this before merging anything to the master branch.

tvatter:
- OK
- Then why about writing the typedef inside the class declaration?
- OK
- OK

PR #11, opened 2017-01-08
Kernel estimator class

tnagler:
I ported most of the C++ functionality from the kdecopula. 
   - A class `InterpolationGrid`: stores gridpoints and function values on a mxm grid; provides functions for interpolating and integrating along one of the axes.
   - A class `KernelBicop`: data-member of class `InterpolationGrid`, uses interpolation to define `pdf1_default()`, `hfunc1_default()`, `hfunc2_default()`, etc. Uses `hinvX_num()` for inverse h-functions. It has a virtual method `fit()` that must be implemented for each kernel estimator.
  - A class `TrafokernelBicop` that implements the basic transformation estimator (`method = "T"` in kdecopula). So far, the bandwidth matrix is selected automatically and there is no way to adjust it.

I think that's sufficient for now. I'll add more estimators and functionality when the vine infrastructure is done.


tvatter:
- There are naming inconsistencies in the `bicop_interpolation.xpp` files. To be consistent with the other stuff, we should either use `BicopInterpolation` for the class name, or use `interpolation_grid.xpp` for the file names and `VINECOPULIB_INTERPOLATION_GRID_HPP`. The first solution is not ideal because `Interpolation` is not a sub-class of `Bicop`. The second solution is not ideal because the code for grid interpolation would not appear directly along with the bicop stuff in the folders. However, I find this latter issue to be the lesser of two evils.
- As a general remark: is is necessary to use the same grid points for both dimensions?
- In `bicop_interpolation.cpp` lines 61-67, wouldn't it be more efficient to start from the end and break out of the loop once `i` and `j` are assigned?
- In `bicop_interpolation.xpp`, you sometimes use `const MatXd& uev`, sometimes `const MatXd& x`, maybe we could use `const MatXd& u` all the time as we did in the `bicop_class.xpp`?
- In `bicop_interpolation.cpp` lines 112-113, you comment parameters that are not present in the function's definition.
- Concerning the naming again, are you sure about `BicopKernel` (compared to e.g. `BicopNonParametric`)?
- Includes in `bicop_kernel.cpp` should probably be in `bicop_class.hpp`. In fact, `gsl_randist` and `gsl_cdf` are already in `bicop_gauss.hpp` and `bicop_student.hpp` and should be also removed from there. Concerning `exception`, it is currently in `bicop_class.cpp` and should also probably be moved to `bicop_class.hpp`... And similar comment for the includes in `bicop_trafokernel.cpp`.
- Since you include `gsl_cdf`, maybe hardcoding the values in `KernelBicop()` isn't the smartest thing to do (but I can live with that). And similar comment for `TrafokernelBicop::fit`.
- Since you include `gsl_randist`, is there a reason for reimplementing the gaussian pdf in `gaussian_kernel_1d`? Why not using `gsl_ran_ugaussian_pdf`? An similarly for `gaussian_kernel_2d`? Why not using `gsl_ran_bivariate_gaussian_pdf`?
- If you define `npars_` as a data member of `BicopKernel`, why do you implement `calculate_npars` at the `Trafokernel`level?

tnagler:
1. You're right, we should go with `interpolation_grid.xpp`.

1. It's not necessary, but there's no reason to do something else. 

1. I agree on breaking the loop, but why is starting from the end more efficient?

1. I used usually use `x` for general real numbers and `u` if it is assumed to be in [0, 1]. Most of the interpolation functionality works in the general case, the `integrate_1d()` assumes input to be in [0, 1].  But the `uev` should be `u`.

1. That's a relict from kdecopula, I'll remove it.

1. Yes, I will only use interpolation for kernel estimators. There's no point in using it for Bernstein and B-spline estimators.

1. Agreed.

1. It was just easier to not deal with it ;) I'll change it.

1. The difference with the `gaussian_kernel_1d` is the truncation at 5. My experience was that truncating the kernels makes the evaluation of the estimator a lot faster. This may not be an issue anymore because we only do it once and then use interpolation. I would leave it this way for now and maybe check the effect of using truncated/non-truncated kernels when other estimators are implemented.

1. Oh yeah that doesn't make sense any longer (previous version had the actual calculation in `TrafoKernel::calculate_npars`.

tvatter:
3. I was not thinking straight. Breaking the loop is enough. Starting from the end should be the same (unless x(n, 0) and x(n, 1) are "large").
4. OK for `u` instead of `uev` then. What about the `data` that is used in our fit rountines? Should we transform `data` into `u`?
9. Can't you still use the truncation along with `gsl_ran_ugaussian_pdf` replacing the righ-hand-side of the equation at line 36?

tnagler:
4. Not sure about changing `data` to `u`, because they are conceptually different. `u` are evaluation points and `data` are data points.

5. Ah, now I know what you mean, I'll change that.

PR #16, opened 2017-01-13
Bicop boost

tvatter:
- Remove GSL dependencies and switch to Boost
- I left the findGSL and the linking to GSL in the findDependencies.cmake in case we want to revert
- An alternative implementation of dpqnorm and dpqt using GSL is commented out in boost_tools.hpp


tnagler:
Approved!

PR #17, opened 2017-01-14
Vine matrix utilities

tnagler:
I created a class `RVineMatrix`. It contains methods for computing some derived matrices required in the estimation and evaluation algorithms (`reorderRVineMatrix`, `createMaxMat`, `neededCondDistr` in VineCopula).


tvatter:
Impressive work... you had a productive week-end ;)

Just a question about design of the `RVineMatrix` class: why are `no_matrix_`, `max_matrix_`, `needed_hfunc1_` and `needed_hfunc2_` data members?

I would avoid storing them an compute them whenever needed (unless their computation is really expensive, but I don't think that's the case).

In my opinion, the code would be much neater if we just made the matrices accessible through getters that would call the methods `to_natural_order `, `to_max_matrix `, `compute_needed_hfunc1 `, `compute_needed_hfunc12` (or simply remove the four functions from the header and use them as definitions of the getters).

PR #25, opened 2017-01-20
Vinecop simulate

tnagler:
So here's the simulation. I don't think there's way to vectorize this, since any combination of rows and columns (to the left) of direct/indirect have to be accessible.

### DIFF
--- a/include/bicop_class.hpp
+++ b/include/bicop_class.hpp
@@ -20,10 +20,9 @@ along with vinecopulib.  If not, see <http://www.gnu.org/licenses/>.
 #ifndef VINECOPULIB_BICOP_CLASS_HPP
 #define VINECOPULIB_BICOP_CLASS_HPP
 
-#include "boost_tools.hpp"
+#include "distribution_tools.hpp"
 #include <Eigen/Dense>
 #include <nlopt.hpp>
-#include <random>
 #include <iostream>
 #include <memory>
 #include <vector>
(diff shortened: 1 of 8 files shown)


tvatter:
1. `invert_order` should be declared in `vinecop_class.hpp` after the class definition.
2. Template functions should be defined in the header (in this case, `to_upper_tri`). Now that I think about it, we should also similarly move the definition of `is_member` (currently declared in `bicop_class.hpp` and defined in `bicop_class.cpp`) in the header.
3. Since the code
```
std::random_device rd;
std::default_random_engine generator(rd());
std::uniform_real_distribution<double> distribution(0.0, 1.0);
auto runif = [&](double) { return distribution(generator); };
```
is copy-pasted, it would be worth transforming into a function. A solution would be to change `boost_tools.hpp` into `distributions_tools.hpp` (or similar) and put all the univariate distribution stuff there.
4. Are you sure that we can't find a way to vectorize `Vinecop::simulate`? Maybe there is a way to find an alternative simulation algorithm, I need to think about it.

tnagler:
1. Will do
2. Will move all the things.
3. I thought about that as well and ended up not writing a separate function because of the `boost_tools` name ;) I'll change the header name and write a separate function.
4. I thought about it a lot and I don't think there's a (reasonable) solution. The only way I could think of is to compute the long chain of transformations for each variable without keeping intermediate results. But this would mean that we compute several hfunctions and inverses several times. This will certainly be slower than the current version.

tnagler:
4. So I came up with a work-around. The implementation is now vectorized over the observations. To avoid exceeding the memory, we estimate the required memory before creating the `direct`/`indirect` arrays. If more than 1GB is required, we simulate `n / 2` observations two times. If `n / 2` still requires too much, we split again, etc.

PR #26, opened 2017-01-21
Hinv speedup

tnagler:
It really annoyed me that our numerical inversion for the h-function was so slow. It only affected the Frank copula so far, but we will probably use it for most family that we add in the future. I did some informal profiling and noticed a substantial overhead from calling the vectorized function `hfuncX_default` on a single point. 

So I implemented a vectorized version of the bisection method that always runs through 35 iterations for all evaluation points. This guarantees an accuracy of more than  `1e-10` on each evaluation. The code is shorter, more elegant, and runs much faster:

Timings on my computer for evaluating Frank's `hinv1` on `1e4` points:
- VineCopula: 85ms
- vinecopulib (old): 260ms
- vinecopulib (new): 37ms

### DIFF
--- a/include/bicop_class.hpp
+++ b/include/bicop_class.hpp
@@ -29,6 +29,7 @@ along with vinecopulib.  If not, see <http://www.gnu.org/licenses/>.
 #include <vector>
 #include <exception>
 #include <cmath>
+#include <functional>
 
 typedef Eigen::VectorXd VecXd;
 typedef Eigen::MatrixXd MatXd;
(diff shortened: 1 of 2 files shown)


tvatter:
I really like this new implementation. However, since `hinv1_num` and `hinv2_num` are the same (my fault initially), couldn't we have a `numerical_inverse` function that would take a vector and an unary function as arguments and return the numerical inverse of the function? Then the implementation of `hinv1_num` and `hinv2_num` would be straightforward...

tnagler:
Now it's even more short and elegant ;)

tvatter:
Nice! Just one thing: why are you using `hfunc1_default` in `hinv2_num`?

tnagler:
Copy&paste error :/ `hinv2_num` isn't used so far, so the tests passed nevertheless.

PR #36, opened 2017-02-06
Appveyor script + removing __attribute__(unused) for compatibility with MS C++ compiles

slayoo:


slayoo:
Besides merging, this will need setting up the Appveyour web hook on github
This is how it looks on my fork: https://ci.appveyor.com/project/slayoo/vinecopulib/build/1.0.49

tvatter:
1. What is the problem with `__attribute__((unused))` and MS C++ compilers ?
2. I'm sorry, I don't understand your comment about the web hook. I created the Appveyor project linked to `tvatter/vinecopulib`, is there anything else to do ?

slayoo:
1. IIUC, __attribute__ is a GNU extension and MS compilers simply do not support it (http://stackoverflow.com/questions/4226308/msvc-equivalent-of-attribute-warn-unused-result)

2. I've commited a dummy change and it triggered both Travis and Appveyor builds - seems nothing else is needed

3. I would definitely not wait as having the integration even without tests will still warn us if something in the library is not compatible with MS compilers (see point 1!)

tnagler:
3. I agree with @slayoo. Just having the library built on windows is useful even without tests.

tvatter:
1. OK, but then won't we have many compiler warnings ?
2-3. OK

slayoo:
1. Besides removing the __attribute_, I have also removed the argument identifiers - they are inaccessible, so there is no reason to warn anymore. The compilation on Travis went OK without any warnings.

tnagler:
Oh, good to know that this is an option ;)

tvatter:
I just noticed `# TODO!` in `compilerDefOpt.cmake`. Is merging the branch without defining the compiler options for windows really a good idea?

slayoo:
I know nothing about Windows compiler options, hence I thought it's better to use the CMake defaults for now, leaving a TODO in the script

PR #40, opened 2017-02-08
moving away the dependence on nlopt.hpp from user-exposed header

slayoo:
Addresses: https://github.com/tvatter/vinecopulib/issues/39


slayoo:
sorry, the second commit was introduced by mistake! (but it does not do any harm)

PR #41, opened 2017-02-09
fix conversion warnings in Visual Studio

tnagler:
I made the double -> int conversion explicit with `(int) x`


slayoo:
helped, thanks!

BTW, any idea why both Travis and Appveyor fired twice for this PR? In https://github.com/tvatter/vinecopulib/pull/30 there are just single builds, while here there is "pr" and "branch" - seems unnecessary...

tnagler:
Yeah it's because I made the branch directly in this repo, not a fork. Otherwise it would fire once on my fork and once on the PR.

PR #47, opened 2017-02-28
fixing structselect_tools.cpp compilation with gcc (https://forum.kde.org/viewtopic.php?f=74&t=118922)

slayoo:


tnagler:
fixed by 1cd236f0be0136fc68c14a8c3c29080fd39b7491

PR #50, opened 2017-03-02
Python bindings

slayoo:


tnagler:
This should go directly into clean-up, so I changed the base and delete refactor-fam-par now.

PR #52, opened 2017-03-08
Archimedean refactoring

tvatter:


tnagler:
1. `ParBicop::fit()` currently contains a lot of duplicated code. I guess some of it could be modularized to make it more concise (especially nlopt setup and try/catch block) .
2. Could you add a `catch (...) {}` at the end of the try/catch block? When experimenting in R a couple of weeks ago, this was necessary to prevent some uncaught errors (but results were always fine).
3. I would replace `str.compare("something") == 0"` with `str == "something"` everywhere. It's just more natural to read.
4. bicop_class.cpp, line 95: `family_exists` became redundant, you can use `if (!is_member(family_set[j], all_families))`-

tvatter:
1. Yes. Actually, I was pondering about refactoring this one completely. How about a `control` argument for the options to `NLopt`? How about a separate namespace for everything internal structures and functions used in `ParBicop::fit()`?
2. Will do.
3. I didn't know that `str == "something"` was OK. If it is, then I will modify.
4. Will do.

tvatter:
1. Commit b7bfd73 implements such an idea with a refactoring of the `ParBicop::fit`. I think that it's much cleaner now. Still wondering about whether `NLoptControl` should be an argument or just something that we use internally (as implemented currently)....
2. In fact, I'm not sure what you mean, as there is already a `catch (...) {}` at the end of the try/catch block. What am I missing?

tnagler:
1. Much better now! I think it can be improved even further though. Could we make the `nlopt::opt` object part of the new (so far only control) class and add methods for setting the bounds from a `MatXd` input? Then all nlopt-specific functionality would be organized in a single class and we just need to construct it, pass parameter bounds / starting values, and hit optimize (should be 3-4 lines of code). 
2. I meant literally adding `catch(...) {}` as a ["catch all" block](http://en.cppreference.com/w/cpp/language/try_catch).

tvatter:
1. Most of the code is about constructing the bounds and starting values in the two different cases (`itau` vs `mle`). As such, we will save only one line by doing this (i.e., `opt.set_lower_bounds(lb);`and `opt.set_upper_bounds(ub);` will become `opt.set_bounds(bounds);`). I'm not fundamentally against doing that, but I don't really see the benefit.
2. Will do.

tnagler:
Very nice, you can merge after conflicts are resolved.

tvatter:
I solved the address sanitizer issue on my computer and it seems to have worked for the OSX build on travis (see 60448db). However, there seems to be a problem with the linux build, and I have no way of reproducing the error. Can you eventually try on your machine?

tvatter:
OK, so now it's passing on OSX again, but still failing on Linux although I don't know why.

tnagler:
This should work now.

PR #56, opened 2017-03-13
Extended README

tnagler:
So i thought it was time to write a more comprehensive README. Comments/changes/additions are welcome. For the future, I suggest that we update the README along with any change in the library that invalidates some of the content in it.

To-dos:
- [ ] Upload Doxygen documentation somewhere and link to it.
- [ ] Check if code examples actually work ;)


tvatter:
In my opinion, this should/could be part of the clean-up. Is there a reason for having a specific branch for this?

tnagler:
Not really. I just didn't want to interfere with the commit history in the other branch and open a separate space for discussion. But can just merge it into the clean-up branch if you want to.

tvatter:
OK, so yes, if there is no specific reason for this branch, I would prefer that we keep all the clean-up related stuff in the clean-up branch.

Issue #20, opened 2017-01-15
Getting rid of R+Rcpp dependencies


tnagler:
I wrote my unit tests in #18 and #19 by hardcoding input values and VineCopula output - mainly because I was to lazy to deal with the R interface (we will remove it at some point anyway). 

It is a bit ugly, but there's little reason not to do it. The tests catch 99% of the errors the R interface would and can be used even with R interface removed. A cleaner version of this would be to write an R script that writes inputs and results into a file which is loaded from within the tests.

What do you think?

tvatter:
I completely agree that it would be nice to remove the dependencies. However, the R interface helped me a lot when writing the pdf/hfunc/hinv for parametric families to benchmark against VineCopula and I think that it can still be useful in the development phase.

I like the idea of using an R script that writes inputs and outputs into files, but there may be a "size" problem. Currently, we are running 5 tests (pdf/hfunc1-2/hinv1-2, not counting tau/loglik) for 4+3*4 families (4 rotationless and 3 with rotations), i.e. 80 tests, for a sample size of 10'000, which means that the output would be a relatively large file (~15MB on my computer)... So we would need to run the R script and clean-up the files after the test is run (i.e., we would still need the R dependencies). Furthermore, we would need still a way to let C++ pass arguments to the R script (at least as long as we're in the development phase).

tnagler:
Is it really necessary to run the tests with a sample size of 10'000? If there's a bug in the code, it will probably also show up when n = 10. The only thing that such a large sample size adds is to (randomly) test the stability of the C++ implementaiton near the boundaries. But this could be done with a few deterministic points as well.

As for passing arguments to R script. We could write a separate setup file which is loaded from both C++ and the R-script.

tvatter:
Concerning the sample size: I agree that it may not be necessary for the tests of pdf/hfunc1-2/hinv1-2 and that ~100 carefully chosen points may be enough. In fact, I increased the sample size to 10'000 to properly test `bicop_sel`, that is to almost guarantee that the right family was chosen (i.e., the probability of random build failure is small). However, if you find a way to carefully create a "small" sample (for each of the 16 tested families) that is "representative enough" (i.e., such that the the right family is chosen), I'm OK with that!

I also agree with the separate setup file. However, we would still depend on R or am I missing something?

tnagler:
[...]

PR #59, opened 2017-03-15
Vinecopulib namespace

tvatter:
To summarize:
1. The namespace is implemented (see f876357).
2. I suggest that we use the actual names from `Eigen` to avoid confusing users (see f30b340).
3. The other commits are from either the [branch portability](/tvatter/vinecopulib/tree/portability) or related to `BUILD_TESTING` (see #20). I know, it was probably not a smart idea to include them here, but it's late ;)

Concerning 2., I think that it was a beginner's mistake to use "custom types". In fact, I would strongly in favor of using `Eigen::XXX` everywhere and removing `using Eigen::XXX;` from `tools_eigen.hpp`. IMO, people that know C++ won't mind the `Eigen::` that we initially wanted to avoid.


tnagler:
I don't mind `Eigen::MatrixXd`. We can still put a `using namespace Eigen` in the `.cpp` files whenever needed, but don't interfere with the users namespace.

PR #61, opened 2017-03-16
change family numbers to enum class + additional refactoring

tnagler:


tvatter:
1. lt/ut => Is it for lowertail/uppertail ? And what about the effects of rotations?
2. We should probably remove family_name_ and association_direction_ from BicopClass. For association_direction_, I noticed that you already did remove it form the constructors of the concrete classes… For family_names_, writing a method to convert the enum class element into a string would be much cleaner IMO.
3. BicopPtr Bicop::create(BicopFamily family, int rotation) => We should add default arguments.
4. What convention do you use for function declaration (one line or multiple lines)? 
5. In preselect_family, vinecopulib:: in vinecopulib::bicop_families::BB is probably useless since you use the namespace.
6. in test_bicop_select, why are you writing family_set = {true_family}; for the MLE (line 32)?
7. As discussed, I think that ifs should have {}.

tnagler:
1. Yes, although the naming is debatable (as it should not imply upper tail *dependence*, but only that the dependence is higher in the upper tail). It always relates to `rotation = 0`, the behavior for other rotations is implicit.
2. `association_direction_` for sure. I actually initially removed the `family_name_` initially, but then google told me that converting the `enum class` to a string is not straightforward (most suggested to use a more complex construction instead of `enum class`). The easiest way would be to cast the `enum class` as `int` and then do `switch` converting integers to the name. But this is highly error prone, because the integers would change if we change the order of the families within the `enum class` declaration. So I put the `family_name_` back in there because this seems to be most convenient.
3. `rotation` has a default (`0`), for `family` we could use `BicopFamily::Indep`?
4. Right, I'll remove it.
5. That's a relict from debugging, will remove it.
6. I'll add them.

tvatter:
2. I was thinking about a method `std::string get_family_name()` implemented with a switch as we do for `create`.
3. Yes for `family`, and I didn't see it for `rotation`, my bad.

tnagler:
2. That would be possible, but I think it's cleaner to just store it as a class member (since the family name is a property of the copula family).
3. I'll add that.

tvatter:
2. I disagree, (a)  the family name is a property of each member of the enum and we may want to print the name of e.g. BicopFamily::Indep without reference to an IndepBicop object, (b) it is never used except for print, and (c) I don't like polluting objects by adding unnecessary data members. I agree that it's more lines of code to write the switch, but at least it will all be in the same place, and we can easily reuse e.g. `get_family_name` to write `get_family_shortname` or something like that.

I also noticed two more things:
7. Sometimes, there is is_member(rotation_, {90, 270}) and sometimes (rotation_ == 90) | (rotation_ == 270), maybe we should be consistent on this.
8. The elements of the enum class `BicopFamily` sometimes start with uppercase and sometimes are fully in uppercase letters, but it may be cleaner to have everything in lower case (also since the vectors of families are in lowercase letters).

PR #68, opened 2017-03-20
trying to get the tests running on windows (path issue)

slayoo:


tvatter:
Yeah, I tried running the tests by executing test_all.exe, but it doesn't seem to be as simple ^^

Windows is really not my strongest suit!

PR #82, opened 2017-03-23
Bicop wrapper

tvatter:


tnagler:
1. The `parameters` argument in the `Bicop`/`AbstractBicop` constructors and factories should be `const Eigen::MatrixXd&` instead of `Eigen::VectorXd` (for nonparametric families this can be large matrix).
2.  I would find it more natural to have the `select()` code in the `select()` function and not the constructor implementation (currently it's the other way around).
3. The 2- and 3- argument constructors of `Bicop` can be merged into one (with default arguments).
4. The error messages in test_bicop_parameteric should be more informative (what family, what function failed).
5. `Bicop::loglik()` can be written as a one-liner, which also avoids creating an additional vector of density values.
6. `rotation_` (and possibly `family_`) and related functions should be part of `Bicop` not `AbstractBicop`.

PR #115, opened 2017-03-29
update vinecopulib version before applying patchs to master

tvatter:
After merging this branch into dev, dev should be merged to master.


tnagler:
Just as a reminder: We should run `make doc` again immediately before merging (in case additional things change).

PR #117, opened 2017-03-29
installing git_revision.hpp

slayoo:


tnagler:
Seems to be causing the windows builds to fail. Any idea?

slayoo:
I'm looking into fixing it right now

slayoo:
fixed

PR #121, opened 2017-03-29
Fix windows warnings

tnagler:
I fixed all warnings caused by our code by using `size_t` throught. This changes the API for Vinecop models, because we no longer use `Eigen::MatrixXi`, but `Eigen::Matrix<size_t, Eigen::Dynamic, Eigen::Dynamic>`.


tvatter:
I think that you should probably update examples/vinecop/src/main.cpp in a similar fashion

PR #126, opened 2017-03-30
fixing library location in Config.cmake.in

slayoo:
was not caught in vinecopulib tests as the actual location /usr/local was part of the default path on Linux/OSX, while on Windows we do not build the example program yet. The pyvinecopulib build failed due to it, though

### DIFF
--- a/cmake/templates/Config.cmake.in
+++ b/cmake/templates/Config.cmake.in
@@ -1,6 +1,6 @@
 @PACKAGE_INIT@
 
-set_and_check(VINECOPULIB_INCLUDE_DIR "@PACKAGE_include_install_dir@/vinecopulib")
+set_and_check(VINECOPULIB_INCLUDE_DIR "@PACKAGE_include_install_dir@")
 include("${CMAKE_CURRENT_LIST_DIR}/@targets_export_name@.cmake")
 check_required_components("@PROJECT_NAME@")
 set(VINECOPULIB_LIBRARIES vinecopulib)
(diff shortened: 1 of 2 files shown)


tvatter:
I do not understand what you want to achieve by removing "/vinecopulib" from `Config.cmake.in`. We wanted to put all the headers in a subfolder /vinecopulib of e.g. /usr/local in order to avoid "polluting" it...

slayoo:
the point is that when you #include <vinecopulib/foo/bar/hpp>, the"vinecopulib" prefix is already there

tvatter:
Yes, but vinecopulib is missing from all the included headers.

slayoo:
Let's raname the "include" folder to "vinecopulib", to enable using #include <vinecopulib/foo/bar.hpp> within vinecopulib codebase

tvatter:
Is it standard and something that "people" do?

slayoo:
Boost does it for example (i.e., they put all includes in the "boost" directory in the release tarball, and within their includes they always use #include <boost/bla/bla...>).

slayoo:
Eigen also has an "Eigen" directory, but they just use #include "Dense" within the internal headers.

tnagler:
how about we put all the stuff in our `include/` into `include/vinecopulib`? don't really like the renaming.

slayoo:
perfect

slayoo:
shall I do it as a part of this PR?

tvatter:
That would be great!

slayoo:
Here it is: https://github.com/vinecopulib/vinecopulib/pull/132/files

PR #132, opened 2017-03-30
moving includes into vinecopulib subfolder of include

slayoo:


tnagler:
That's weird. It worked in Debug and Release modes, but not with `cmake ..` (for OSX).

tnagler:
Thanks!

Issue #124, opened 2017-03-30
0.0.1 tarball does not ship with git_revision.hpp -> "make install" will fail


slayoo:

tvatter:
In windows?

slayoo:
on all platforms

tvatter:
I thought that we wanted to exclude git_revision.hpp from the commits, so I added it to gitignore.

What should we do?

tvatter:
And with the current version, "make install" is working on my machine btw...

slayoo:
Yes, the git_revision.hpp would best be excluded from commits, but it needs to be shipped in the .zip and .tar.gz release files to make "make install" be able to copy it to system directories for users that didn't use git to obtain the library.

I suggest just adding these files to the .zip and .tar.gz release archives.

tnagler:
I think the the source .zip and .tar.gz are actually created automatically from github.

tnagler:
And isn't the fit revision file created automatically by CMake?

tvatter:
Yes, but only if it finds the .git folder, which is not included in the tarball...

tvatter:
As said before, I'm against using this "hack" to manage the versions. I think that even in the age of git, we should use the master branch for pyvinecopulib and rvinecopulib, and control releases to master "manually" by increasing the major/minor/patch number whenever appropriate.

slayoo:
For the record, I'm just suggesting things, and will not take part in the decision processes here :)

tvatter:
Well, if Thomas and you think that we should keep using it, I'll gladly oblige, but if the decision was only mine,  I would have removed it long ago ;)

tnagler:
OK, since this is causing us so much pain and we have actual versions we can refer to now, I think we can remove the revision thing.

slayoo:
I'll comment it out tomorrow

slayoo:
https://github.com/vinecopulib/vinecopulib/pull/135

tnagler:
fixed by #135

PR #135, opened 2017-03-30
commenting out git_revision.hpp logic

slayoo:


tvatter:
Would you mind if I push something to remove it completely?

I would rather avoid just "commenting things out" if we push the patch to master to solve #124, and this logic will still be in the github history if we decide to revert later on...

PR #150, opened 2017-04-04
build an example program in Appveyor (aimed to test the installation)

slayoo:


tvatter:
So, in this one, we're building but still not running the example program, correct?

slayoo:
correct - this is just aimed at testing the installation
execution is commented out as it does not work :(

PR #167, opened 2017-04-19
fixing compatibility issue with (buggy) CMake 3.8

slayoo:
see https://gitlab.kitware.com/cmake/cmake/issues/16816


tvatter:
Why were the backslashes converted into forward slashes only for DBOOST_ROOT ? 

Shouldn't we do that for DEIGEN3_INCLUDE_DIR, DCMAKE_LIBRARY_PATH and DCMAKE_INCLUDE_PATH ?

slayoo:
There was a bug (regression) introduced in CMake's FindBoost.cmake between Cmake 3.7.2 and 3.8 - it only concerns Boost (and only Windows). Without this fix, any Appveyor build would fail.

PR #178, opened 2017-05-12
using find_package(NLopt) instead of custom CMake script

slayoo:


tvatter:
Could you remove completely `findNLopt.cmake` from the the cmake folder and add the file's content to `findDependencies.cmake`?

It would be cleaner in my opinion...

tnagler:
I agree.

slayoo:
Re finNLopt.cmake: there is a hacky workaround for GCC 6 there which is why I would suggest to keep it in a separate file

Re head... this is of course true, but in case of nlopt there is one caveat, there have been no new releases of nlopt for 3 years, while the github repo seems to be used for bugfixing rather than new features

tvatter:
Re finNLopt.cmake: OK, but then don't we need the hack for the examples too then?
Re head: OK for me!

slayoo:
in the examples we don't use the SYSTEM qualifier for the external headers, i.e.:

    include_directories(${external_includes})

vs.

    include_directories(SYSTEM ${external_includes})

and the GCC 6 issue is only related with the SYSTEM keyword. Without the hack, the compilation failed on Travis

tvatter:
Okok, and I wrote most of the cmake related stuff, but I'm quite sure that I didn't add `SYSTEM`, so, a last question and then I stop: why are we using `SYSTEM` already?

slayoo:
It silences tons of warnings that the new version of compilers (especially clang) emit when parsing system and external library headers

tvatter:
Re finNLopt: Alright, perfect then!

Re HEAD: can you add something to the README for OSX users before we merge? I would gladly do it, but I can't commit to your branch...

slayoo:
What about: https://github.com/slayoo/vinecopulib/commit/a68cd059a92e7326a83123ebbe847ac89ea7344c

PR #187, opened 2017-05-18
fix outdated url in license statement

tnagler:
via fix&replace all


tvatter:
Shouldn't you push to the docs branch?

tnagler:
yeah why not

PR #189, opened 2017-05-25
Vinecop cdf

tvatter:


tnagler:
Looks pretty good, just three minor things:
1. `ghalton()` should be documented with including a reference.
2. You sometimes write "haston" instead of "halton".
3. Since it is used a couple of times, 
```
if (u1 <= u(i,0) && u2 <= u(i,1)) {
    return 1.0;
} else {
    return 0.0;
}
```
could be a bit more compact, e.g., `(u1 <= u(i,0) && u2 <= u(i,1)) ? 1.0 : 0.0` or `(double) (u1 <= u(i,0) && u2 <= u(i,1))` (not sure if second works).

Issue #186, opened 2017-05-18
Checks for the validity of an R-vine matrix


tnagler:
... should be added to the `RVineMatrix` class at some point.

tvatter:
This is extremely weird, because the select function remove edges that are not allowed by the proximity condition. As such, constructors from data where no RVineMatrix is provided by the user should not return such a matrix.

However, the feature is indeed very important for user-provided matrices.

tnagler:
Added checks in #192. 

@bodashtart: The checks are run whenever `RVineMatrix()` is called, in particular before `Vinecop(data)` returns the result. If there's an error on the C++ side, you will see an error message.

tvatter:
Can we close it then?

tnagler:
Yes, all fixed.

PR #192, opened 2017-06-02
Check for validity of R-vine matrices

tnagler:
re #186 

Added checks for several conditions that must be satisfied by an R-vine matrix. The checks are actually redundant, `check_upper_tri()`, `check_antidiagonal`, `check_columns()` are violated whenever `check_proximity_condition()` is, but I kept them to allow for more informative error messages. They could be ommited for sake of performance, which I haven't looked into so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,6 +468,23 @@ Denoting by `M[i][j]` the matrix entry in row `i` and column `j` (starting at
 2. Jump up to the element in row `t` (second conditioned variable).
 3. Gather all entries further up in column `e` (conditioning set).
 
+A valid R-vine matrix must satisfy several conditions which are checked
+when `RVineMatrix()` is called:
+1. The lower right triangle must only contain zeros.
+2. The upper left triangle can only contain numbers between 1 and d.
+3. The antidiagonal must contain the numbers 1, ..., d.
+4. The antidiagonal entry of a column must not be contained in any 
+   column further to the right.
+5. The entries of a column must be contained in all columns to the left.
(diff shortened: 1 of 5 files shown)


tvatter:
While I find this version with "one function per check" very intuitive, it is also a bit naïve and I am concerned about performance:

1. Could you bring the loop over the trees in `check_matrix`?
2. In `rvinecopulib`, since we are creating a new objects all the time, we'll need to either accept a huge overhead (from the double loop in `check_proximity_condition`) or find a way to bypass the `check_matrix`. I would be in favor of the later, and a simple solution would be a `safe` argument to the `RVineMatrix`'s constructor.

tnagler:
1. Since loops have almost no overhead in C++, running the same loop several times shouldn't be a problem.
2. Yes, we should only check the validity of the matrix only once. We could write an R-class equivalent to `RVineMatrix` which does the check and must be used when constructing a `vinecop_dist`. When an `RVineMatrix`-equivalent object is passed to cpp, we should not repeat the checks. This will be easy to do with a flag in the constructor. We should also omit the checks at the end of `Vinecop(data)` as soon we are sure that there is no flaw in our logic.

tnagler:
1. Checked with a single loop in dimensions from 10 to 1000. There is no difference between a single loop and several loops in sequence. I'd prefer to keep the modularized version then, the code is much cleaner.

tvatter:
1. If there is no difference in performance, then I completely agree.
2. Exactly what I had in mind!

PR #194, opened 2017-06-06
Flip + get/set_parameters for kernel bicop

tvatter:
Note: unit tests are coming but I'll finish them tomorrow morning!

Before merging to dev  (and then releasing 0.0.3 to master), we should probably discuss the organization of the documentation.

I noticed that the non-virtual method of AbstractBicop were documented, but the non-virtual methods of derived classes (e.g. ParBicop and KernelBicop) were not. In my opinion, we should either document all of them or none. 

Also, looking at the current documentation (e.g. https://vinecopulib.github.io/vinecopulib/classvinecopulib_1_1_abstract_bicop.html), I think that it's a bit messy (what is documented and what isn't).

### DIFF
--- a/include/vinecopulib/bicop/abstract.hpp
+++ b/include/vinecopulib/bicop/abstract.hpp
@@ -14,20 +14,6 @@
 
 namespace vinecopulib
 {
-    // Pre-declaration to allow AbtractBicop to befriend the two functions
-    namespace tools_optimization
-    {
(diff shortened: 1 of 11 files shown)


tnagler:
I think our reasoning was to not document stuff that is not user-facing. The only reason this stuff is in the API documentation is the diagram.

PR #198, opened 2017-06-07
allow to omit R-vine matrix check in Vinecop ctors

tnagler:


tvatter:
Wouldn't it also be nice to use this PR to add the ctors allowing to pass a RVineMatrix object directly?

tnagler:
Do we need such constructors?

tvatter:
They are quite natural IMO: why should we only allow the user to pass a matrix? Currently, the RVineMatrix is documented and part of the user-facing API...

tvatter:
Anyway, we can always add this later.

PR #205, opened 2017-06-12
Serialization

tvatter:
So, it's working fine for both bicops & vinecops on linux & os x, and also for bicops on windows, but for some reason there is an issue for vinecops on windows. Since the error message from appveyor is uninformative as hell, I have absolutely no idea what's wrong...

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,29 @@ auto aic  = bicop.aic(sim_data);
 auto bic  = bicop.bic(sim_data);
 ```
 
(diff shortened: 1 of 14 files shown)


tnagler:
Looks all good. How about adding a small section to the README?

tvatter:
OK! Should I do that + squash/merge?

tnagler:
Yep, and you do not need to wait for the checks again.

Issue #163, opened 2017-04-07
introduce "named parameters" instead of positional arguments in the API?


slayoo:
We have now ctors with up to 9 positional parameters with default values. 
What about switching to something mimicking named parameters from other languages?
For example: https://isocpp.org/wiki/faq/ctors#named-parameter-idiom

tvatter:
I like the idea. How ``standard'' is this particular approach? In https://marcoarena.wordpress.com/2014/12/16/bring-named-parameters-in-modern-cpp/, I noticed many ways to achieve essentially the same goal.

In any case, it seems that this approach is particularly well suited to our FitControlsBicop and FitControlsVinecop classes...

tnagler:
Honestly: I think that's no longer going to happen. Close?

slayoo:
Definitely :)

PR #206, opened 2017-06-12
Sparse selection

tnagler:
Now that everything is working for structure and family selection, I think it's time to discuss what I've done. 

1. Mainly, there was a major refactoring of `tools_structselect.xpp` which is now `tools_select.xpp`.  There's a new class `VinecopSelector` with child classes `FamilySelector` and `StructureSelector`. This allows to use the same high-level code for family and structure selection, where specifics are implemented in the respective child class. 

2. The two functions `Vinecop::select_all()` and  `Vinecop::select_families()` are now essentially duplicates and could be merged into one. 

3. The current implementation has a larger memory footprint than necessary. The R-vine matrix and pair copulas are stored twice (once in `Vinecop` and once in `VinecopSelector`. This is unlikely to be a bottleneck, because both objects are rather small (unless a lot of kernel copulas are stored). This 'issue' could be fixed by storing pointers to `vine_matrix_` and `pair_copulas_` instead of copies, or by making `VinecopSelector` a child class of `Vinecop`.


tvatter:
1. I don't mind, but in this case `tools_bicopselect.xpp` should probably also be renamed `tools_select.xpp`.

2. No strong opinion on this. I thought that having two functions was for user friendliness, but it might actually be more confusing than user-friendly.

3. Wait, but the `VinecopSelector` is destructed right after the data is copied, so the memory issue is not something that will be dragged along or am I missing something?

So, my comments (12 is arguably the most important):

4. Why is `include/git_revision.hpp` reappearing?

5. There is a lot of duplicated code in `cor/tau/hoeffd_matrix`. Could you eventually have a single function, e.g. `dependence_matrix`,  with an additional argument for the type of dependence measure? By the way, I noticed that while looking at the coverage report (i.e., cor/hoeffd are not unit tested).

6. In `vinecop/fit_controls.hpp`, I think that `show_trace` should be the last argument. By the way, the list of arguments gets reaaaaally long, I think that named parameters (see #163) for the arguments of the controls' constructors would really improve the user friendliness.

7 In `vinecop/fit_controls.cpp`, the "new arguments" (`select_truncation_level` and `select_threshold`) are not documented.

8. In `tools_select.hpp`, why did you implement a default constructor and destructor to `VinecopSelector`? Aren't those generated by default? Same question for the destructors of `FamilySelector` and `StructureSelector`.

9. In `tools_select.hpp`, why didn't you indent the content of the `tools_select`'s namespace?

10. In `tools_select.cpp`, line 118, why does the loop starts with `size_t i = 1`?

11. In `tools_select.cpp`, you document a `learning_rate` for `get_next_threshold`, but it does not appear anywhere.

12. OK, I'm stopping the detailed comments and taking a step back. I think that there is a lot of "duplicated" stuff (e.g. all the gets/sets/print) due to the fact that you are using a "different infrastructure" for `FamilySelector` and `StructureSelector`. Wouldn't it be possible to use `std::vector<VineTree> trees_;` in `FamilySelector` too? For instance, by having an alternative `add_allowed_edges` for `FamilySelector` and with other minor modifications, it seems to me that the same infrastructure could be used in both cases. In my opinion, it would make the code much easier to read, maintain and extend.

tnagler:
1. Good idea.
2. It was, but at a time where we had no `FitControls` and the arguments got confusing when mixing both functions. Now it would be just `(data, controls, matrix = Eigen::Matrix<...>())`. So I would vote for a single function, but also don't have a strong opinion.
3. No, nothing is dragged along. But while the fit is running, the function requires more memory than necessary. It's not user-facing, so I would leave it as is for now. We can optimize later in case we find it to be problematic in benchmarks. 
4. Should not appear, probably due to the fact that I started working on it 2.5 months ago. I'll remove it.
5. Good idea!
6. Yeah I'm also leaning more and more towards named arguments. But that's not something we should do within this PR.
7. I'll add documentation.
8. Thank's for the reminder. I was just adding those in the first draft, because I remembered that there was something about destructors in abstract classes. It never caused problems, so I forgot to think about it further. After a quick google search, it seems that the `VinecopSelector` destructor [should actually be virtual](https://stackoverflow.com/questions/461203/when-to-use-virtual-destructors) which requires the derived classes' destructors to be defined.
9. That's a leftover from a previous version where I had two sub-namespaces within tools_select. I'll indent it properly.
10. In order to exclude the diagonal which is always 1 (and does not correspond to a proper pair-wise dependence measure).
11. Yeah, I had that argument in an earlier version. It's `alpha` now and defined directly within `get_next_threshold()`. I'll need to revise all documentation anyway, just wanted to discuss the implementation before working too hard on this.
12. I noticed this as well. The reason is mainly, that using the boost graph infrastructure feels a bit over the top when all algorithms can easily be written with just the matrices. It's certainly easier to understand what's going on in the family selection, but I agree that having two different infrastructures can be confusing as well. It's certainly possible to do everything with the boost graph infrastructure. I need to think about it a little more.

tvatter:
2. Agreed.
3. Exactly, let's worry about that when benchmarking.
6. OK, but we still need this refactoring to start with the name arguments.
12. Yeah, I agree that the graph infrastructure would be overkill if it was used just for that. However, since it is used for the other selector, then someone would still need to understand this part of the code. Again, correct me if I'm wrong, the main difference would be  an alternative `add_allowed_edges` function (and potentially skipping the MST, although if it's already a tree, it may not even be necessary).

tnagler:
Just wanted to do 2. and realized that `select_families()` doesn't have a matrix argument, it uses the `vine_matrix_` of the `Vinecop` object. So the two functions are actually needed to tell whether or not the structure should be selected.

tvatter:
Just two questions: 

1. Why did you keep `cor_matrix` and `hoeffd_matrix` in the .hpp?
2. If we need the destructor in `VinecopSelector`, don't we also need it in `AbstractBicop`?

tnagler:
1. unintentional, removed them.
2. #207

PR #215, opened 2017-08-20
normalization of margins for kernel estimators

tnagler:


tvatter:
Just to be sure to understand what's going on: is there a reason for doing rows first and then columns?

tnagler:
No that's an arbitrary choice.

PR #217, opened 2017-08-22
Refactor sparse select

tvatter:


src/vinecop/tools_select.cpp
```diff
@@ -453,40 +447,40 @@ namespace tools_select {
...
+            mat(t, col) = trees_[t][e0].conditioned[0];
+            mat(t - 1, col) = trees_[t][e0].conditioned[1];
     
             // assign fitted pair copula to appropriate entry, see
             // vinecopulib::Vinecop::get_pair_copula().
             pair_copulas_[t - 1][col] = trees_[t][e0].pair_copula;
     
             // initialize running set with full conditioing set of this edge
-            auto ned_set = trees_[t][e0].conditioned;
+            auto ned_set = trees_[t][e0].conditioning;
```
tnagler: should be called `ning_set` then ;)

src/vinecop/rvine_matrix.cpp
```diff
@@ -83,7 +95,6 @@ namespace vinecopulib
         std::vector<size_t> conditioning_test(tree);
         std::vector<size_t> conditioned_test(2);
         bool res = false;
```
tnagler: Same as below, `res` could have a more expressive name like `is_edge` or `is_in_structure`.

PR #218, opened 2017-08-22
Refactor sparse select fix

tnagler:
- Fixed the rotation issue in `FamilySelector`: your `set_sym_diff` "forgot" about the ordering of the sets, which is fixed now. This makes the `v0_` thing superfluous and allows to simplify `belong_to_structure()`.
- Simplified other functions like `add_edge_info()` a bit by storing `all_indices` in vertices as well.
- Simplified `finalize()` for `FamilySelector`: since the matrix is fixed, we need to do a lot less work. `finalize()` is now virtual in `VinecopSelector` and each child class has their own implementation.
- Did all the stylistic changes that I suggested in #217.

We should keep the commits when merging into `refactor-sparse-select` and only squash when merging to `dev`.

### DIFF
--- a/include/vinecopulib/misc/tools_stl.hpp
+++ b/include/vinecopulib/misc/tools_stl.hpp
@@ -52,21 +52,6 @@ namespace tools_stl {
     }
 
     template<class T>
-    vector<T> set_sym_diff(vector<T> x, vector<T> y)
-    {
-        sort(x.begin(), x.end());
-        sort(y.begin(), y.end());
-        vector<T> different;
-        set_symmetric_difference(
-                x.begin(), x.end(),
-                y.begin(), y.end(),
-                back_inserter(different)
-        );
-
-        return different;
-    }
-
-    template<class T>
     size_t find_position(T x, vector<T> vec)
     {
         return distance(vec.begin(), find(vec.begin(), vec.end(), x));
@@ -103,6 +88,14 @@ namespace tools_stl {
         out.reserve(1 + y.size());
         out.insert(out.end(), y.begin(), y.end());
         return out;
+    }
+    
+    template<class T>
+    vector<T> set_sym_diff(vector<T> x, vector<T> y)
+    {
+        vector<T> dxy = set_diff(x, y);
+        auto dyx = set_diff(y, x);
+        return cat(dxy, dyx);
     }
 
     template<class T>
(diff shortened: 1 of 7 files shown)


include/vinecopulib/misc/tools_stl.hpp
```diff
@@ -104,6 +89,14 @@ namespace tools_stl {
         out.insert(out.end(), y.begin(), y.end());
         return out;
     }
+
```
tvatter: I think that mine was more efficient (e.g., the sort is only once)...

src/vinecop/tools_select.cpp
```diff
@@ -298,12 +298,81 @@ namespace tools_select {
...
-                                             const RVineMatrix& vine_matrix) {
-
-        bool res = false;
+    bool FamilySelector::belongs_to_structure(size_t v0, size_t v1,
+                                              const VineTree& vine_tree) {
+        
+        // -1 means no common neighbor in previous tree
         if (find_common_neighbor(v0, v1, vine_tree) > -1) {
             // conditioned sets
             auto conditioned0 = vine_tree[v0].conditioned;
```
tvatter: Wouldn't it be "nicer" to use all_indices rather than conditioned/conditioning here? We would avoid the additional `cat`...

include/vinecopulib/misc/tools_stl.hpp
tnagler: Yes but [your result was sorted](http://en.cppreference.com/w/cpp/algorithm/set_symmetric_difference) as well. This makes it impossible to check whether conditioned sets in the tree representation are exactly equal to the ones implied by the matrix.

tvatter: But the two separate set_diffs are sorting the vector too aren't they?

tnagler: But only before `cat()` is called. The order of the result is not sorted.

src/vinecop/tools_select.cpp
tnagler: Agreed.

include/vinecopulib/misc/tools_stl.hpp
tnagler: 4ce5622089ea5e63d4ef09695f6cdbf1dd4502cf

src/vinecop/tools_select.cpp
tnagler: 4ce5622089ea5e63d4ef09695f6cdbf1dd4502cf

include/vinecopulib/misc/tools_stl.hpp
tvatter: I see!

PR #241, opened 2017-10-24
Remove nlopt mem leak

tnagler:


tvatter:
If the optimization failed, can't the loop lines 196-198 also fail? In this case, it seems unwise to wait after the loop to throw the error doesn't it?

tnagler:
I don't see how this could fail, since we know for sure that
- `optimized_parameters` is a vector of size `n_parameters_`
- `x` is a pointer array of size `n_parameters_`
- there are no issues when I comment out the optimization (lines 173-192)

So the only issue I can imagine is that the optimizer calls delete on `x` or one of its entries, but then we have a much larger problem ;)

tnagler:
We could also just put the loop inside the try/catch.

PR #246, opened 2017-10-25
Header only

tvatter:
It's weird, the library compiles and the tests run fine. However, there is an issue with the parallel stuff when compiling the example. 053913a is supposed to fix this, but I wonder how it could have worked before...


tvatter:
So, it looks like the x64 is failing for the same reason as this one:

https://ci.appveyor.com/project/vinecopulib/vinecopulib/build/1.0.621

Have you actually witnessed an x64 build pass the tests with the parallel stuff?

tnagler:
Among others, the builds for the actual parallelization PR passed: https://ci.appveyor.com/project/vinecopulib/vinecopulib/build/1.0.609

PR #259, opened 2017-11-06
parallelize whenever possible even more

tvatter:


tvatter:
I don't think that having `#ifndef INTERFACED_FROM_R` also in `include/vinecopulib/misc/tools_parallel.hpp` is a good idea. Is it necessary?

tnagler:
If we want to have `map_on_pool()` in the same file, I think it is.

tvatter:
Then I would take the Threads stuff in a separate file and get e.g. tools_threads.

Also, 1e-2 seems like a big difference if the models are equals no?

tnagler:
OK, will do the tools_threads then.

I had `EXPECT_DOUBLE_EQ()` before and it failed on the 8th digit or something. The log likelihood is around 900, so I don't think 2 digits is really that large. Didn't want to spend too long figuring out how far we can go, just do see random failures again ;)

tvatter:
Perfect then!

tnagler:
Kept the ThreadPool in tools_parallel, can also make a new namespace tools_thread if you prefer.

tvatter:
Nah, it's fine as it is now, my suggestion was mostly to do avoid having the R-stuff hidden by the Threadpool implementation.

tnagler:
Yeah makes sense.

PR #263, opened 2017-11-07
Fix Frank and Joe itau methods

tnagler:


tvatter:
Maybe we should use the lower/upper bounds to avoid this kind of issue if we decide to change again in the future?

tvatter:
Also, why did you add `#include <boost/math/tools/minima.hpp>`?

tnagler:
both fixed

tvatter:
All good then!

PR #266, opened 2017-11-08
allow 'loglik' as selection criterion

tnagler:


tvatter:
Didn't you mention that this was the mother of all bad ideas?

tnagler:
Yeah, but I don't think we should disallow it. For example, Dani's  selection algorithm for D-vine regression is based on log-likelihoods (with option for AIC/BIC correction). There may be other use cases that we don't see right now. In any case, there's really no need to patronize the users if it doesn't cost us anything to keep it more flexible.

tvatter:
OK, no problem. In any case, I noticed that you removed the proper casts that are required for Ripley!

tnagler:
Oh damn, seems like I branched out of a wrong branch. I'll just make a new PR.

tvatter:
Yeah just noticed

PR #267, opened 2017-11-08
allow 'loglik' as selection criterion

tnagler:


tvatter:
The README and should be updated as well. Otherwise it can be merged.

PR #299, opened 2018-01-31
Dlls go to prefix/bin

jschueller:
This is the standard on cygwin/mingw/msvc


tvatter:
Thanks a lot, neither Thomas nor I are windows specialists, so contributions are always welcome.

However, would you mind rebasing your PR to the dev branch?

We do not accept PRs to master directly (we should write this somewhere).

PR #304, opened 2018-02-14
mBICV

tnagler:
short summary to guide you:
- threshold and truncation level are now selected by mBICV.
- `FitControlsVinecop` has an option to use mBICV as `selection_criterion = "mbicv"` for family selection.
- this is implemented by adding a `selection_criterion = "mbic"` (note the missing "v") in `FitControlsBicop`.
- both `"mbic"` and `"mbicv"` are parameterized by an additional fit control `psi0`, which is the prior probability for non-independence in the first tree.
- new functions `Bicop::mbic()` and `Vinecop::mibcv()`.
- `Vinecop` has a new data member `threshold` (incl getter) which is 0 by default.


include/vinecopulib/bicop/class.hpp
```diff
@@ -86,6 +86,9 @@ class Bicop
 
     double bic(const Eigen::Matrix<double, Eigen::Dynamic, 2> &u) const;
 
+    double mbic(const Eigen::Matrix<double, Eigen::Dynamic, 2> &u, 
+                const double pi) const;
```
tvatter: should be psi0

tvatter: or am I missing something?

include/vinecopulib/bicop/implementation/fit_controls.ipp
```diff
@@ -96,9 +100,19 @@ FitControlsBicop::check_nonparametric_mult(double nonparametric_mult)
...
         throw std::runtime_error(
-            "selection_criterion should be 'loglik', 'aic', or 'bic'");
+            "selection_criterion should be 'loglik', 'aic', 'bic', or 'mbic'");
+    }
+}
+
+inline void
+FitControlsBicop::check_psi0(double psi0)
+{
+    if (!(psi0 > 0.0) | !(psi0 < 1.0)) {
```
tvatter: psi0 = 1 would mean that we force non-indep, and psi0 = 0 the opposite, right? OK it may not be so useful.

include/vinecopulib/vinecop/fit_controls.hpp
```diff
@@ -27,6 +27,7 @@ class FitControlsVinecop : public FitControlsBicop
                        std::string tree_criterion = "tau",
                        double threshold = 0.0,
                        std::string selection_criterion = "bic",
+                       double psi0 = 0.9,
```
tvatter: Haha, so you think that this one will be used more often than select_truncation_level/threshold ?

include/vinecopulib/bicop/class.hpp
tnagler: oh right, that's debatable. psi0 is really meant to be the "base prior" in the mBICV and "pi" just the prior for a single pair. but I agree that it's inconsistent

include/vinecopulib/bicop/implementation/fit_controls.ipp
tnagler: yeah we have `family_set` and `selection_criterion = "loglik"` for that.

include/vinecopulib/vinecop/fit_controls.hpp
tnagler: no, I didn't really think about the positioning. should we put it before `num_threads`?

tvatter: don't know actually, no strong opinion on this, and actually wondering the same about the threshold

include/vinecopulib/bicop/class.hpp
tvatter: yeah then I would just avoid pi, I don't really like to you pi for anything else than pi

tvatter: (Bayesians are weird)

include/vinecopulib/bicop/implementation/fit_controls.ipp
tvatter: k

include/vinecopulib/vinecop/fit_controls.hpp
tnagler: honestly, I don't think many people will use the constructor with more than the first 1-3 arguments. I always use `.set_xxx()` as soon there are more than 2.

include/vinecopulib/bicop/class.hpp
tnagler: yeah, I'll change to psi0, got caught up too much in those weird Bayesian papers ;)

PR #319, opened 2018-04-26
Add getters for Kendall's tau

tnagler:


tvatter:
I'm all good with this one as soon as the tests pass.

PR #321, opened 2018-04-26
Minor fixes for Eigen tools

tnagler:


tvatter:
That looks good to me: we probably don't need to wait for the checks on this one.

PR #327, opened 2018-05-04
Loglik

tvatter:


tnagler:
shouldn't we rather have a default loglik of `nan` and throw an error if it hasn't been changed to something meaningful?

PR #331, opened 2018-05-04
fix loglik-related bugs [NOT READY TO MERGE]

tvatter:


tvatter:
So, there is still something that I don't like: this flip won't work for models in flip_by_rotation such that the log-likelihood has not been set (i.e., if you try to flip a Clayton with custom initialization).

tnagler:
How about
``` cpp
inline void Bicop::flip()
{
    double ll = bicop_->get_loglik();
    // flipping
    // ...
    bicop_->set_loglik(ll);
}
```
and throwing the error from `Bicop::get_loglik()` rather than `AbstractBicop::get_loglik()`?

tnagler:
Oh, OK that's roughly what you did.

tvatter:
Actually, moving the error throwing is clearly the best idea.

PR #345, opened 2018-06-01
Numerical stability

### LINKED ISSUE #337: log1p overflows
For example:
```r
> vine$pair_copulas[[2]][[1]]
#> Bivariate copula fit ('bicop'): family = bb6, rotation = 0, parameters = 1.26, 1.33
> hbicop(c(0, 0), cond_var = 2, vine$pair_copulas[[2]][[1]], inverse = TRUE)
#> Error in bicop_hinv2_cpp(if_vec_to_matrix(u), bicop) : 
#>   Error in function log1p<double>(double): Overflow Error
```
We probably need to add some safety barriers when doing the generator computations in Archimedean families.

tnagler:
* fixes #337 using a custom `log1p()`
* makes sure that the loglik used in fitting is the same as `pdf(u).log().sum()` by truncating in `AbstractBicop::pdf()` instad of `Bicop::pdf()`

### DIFF
--- a/include/vinecopulib/bicop/implementation/bb6.ipp
+++ b/include/vinecopulib/bicop/implementation/bb6.ipp
@@ -5,9 +5,7 @@
 // vinecopulib or https://vinecopulib.github.io/vinecopulib/.
 
 #include <vinecopulib/misc/tools_integration.hpp>
-
-#include <boost/math/special_functions/expm1.hpp>
-#include <boost/math/special_functions/log1p.hpp>
+#include <vinecopulib/misc/tools_stl.hpp>
 
 namespace vinecopulib {
 inline Bb6Bicop::Bb6Bicop()
@@ -23,13 +21,13 @@ inline Bb6Bicop::Bb6Bicop()
 
 inline double Bb6Bicop::generator(const double &u)
 {
-    double res = boost::math::log1p(-std::pow(1 - u, parameters_(0)));
+    double res = tools_stl::log1p(-std::pow(1 - u, parameters_(0)));
     return std::pow((-1) * res, parameters_(1));
 }
 
 inline double Bb6Bicop::generator_inv(const double &u)
 {
-    double res = boost::math::expm1(-std::pow(u, 1 / parameters_(1)));
+    double res = std::expm1(-std::pow(u, 1 / parameters_(1)));
     return 1 - std::pow(-res, 1 / parameters_(0));
 }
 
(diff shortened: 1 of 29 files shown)


include/vinecopulib/bicop/implementation/bb6.ipp
```diff
@@ -23,21 +21,21 @@ inline Bb6Bicop::Bb6Bicop()
...
 {
-    double res = boost::math::log1p(-std::pow(1 - u, parameters_(0)));
+    double res = tools_stl::log1p(-std::pow(1 - u, parameters_(0)));
     return std::pow((-1) * res, parameters_(1));
 }
 
 inline double Bb6Bicop::generator_inv(const double &u)
 {
-    double res = boost::math::expm1(-std::pow(u, 1 / parameters_(1)));
+    double res = std::expm1(-std::pow(u, 1 / parameters_(1)));
```
tvatter: std::expm1 is better than boost::math::expm1?

tvatter: Same question for stp::log1p btw.

tnagler: I'd always prefer using STL versions when they exist, but no strong opinion on that.

tvatter: Ok, I didn't even know that STL had such functions (just noticed that they were C++11, which probably explains why I didn't find them initially).

I'm all good once the tests pass then.

include/vinecopulib/misc/implementation/tools_interpolation.ipp
```diff
@@ -190,7 +190,7 @@ InterpolationGrid::intergrate_1d(const Eigen::MatrixXd &u,
             tmpgrid.col(0) = grid_points_;
             tmpgrid.col(1) = Eigen::VectorXd::Constant(m, u2);
         }
-        tmpvals = interpolate(tmpgrid);
+        tmpvals = interpolate(tmpgrid).array().max(1e-4);
```
tvatter: 1e-4 is rather "large" no?

tnagler: yes and no.. the integral over [0, 1] would be just 1e-4 which is rather small compared to the actual value of the integral (1). I came up with this value in a few experiments (nothing too solid) that showed it shouldn't be less than `1e-5` because estimates get highly instable (set to `1e-10`) below that.

tvatter: I see, then it's fine. Any idea why the windows check fail?

tnagler: `there is no package called 'mvtnorm'`
I blame ALTREP as always ;) But what confuses me is that the checks run for 1hr, they should be done after ~30mns.

PR #363, opened 2018-07-30
parallelize loglik/bic etc

tnagler:


tvatter:
Shouldn't we also be able to compute the loglik/AIC/BIC from the saved value (e.g., with a default empty `const Eigen::MatrixXd &u`) ?

tnagler:
Shouldn't this rather be `get_aic()` etc?

tvatter:
Agreed, it's just to avoid recomputing in R whenever we can avoid.

tnagler:
So either we add a `nobs` data member, or we can't do `get_bic()`. The "recomputing" in R is really just multiplying two numbers and adding another one, so I think that overhead would be negligible (if it even exists).

tvatter:
Ah true... well, I guess that, if we save the loglik, for a fitted object, it makes sense to save nobs too.

tnagler:
alright added the nobs

tvatter:
Just one thing: shouldn't `compute_mbicv_penalty` be `calculate_mbicv_penalty` or to be consistent with `compute_npars` ?

tnagler:
fixed

PR #367, opened 2018-08-07
implement Rosenblatt transform

tnagler:


tvatter:
Why did you remove the ifelse for the `d > 2` ?

tnagler:
the behavior was inconsistent; no matter if the order was 1-2 or 2-1, the second variable of `u` was kept. noticed when writing the unit test for d=2

PR #371, opened 2018-08-12
fix windows warnings

tnagler:


tvatter:
LOL, I was sure that I had committed this stuff after correcting in rvinecopulib, I just checked and I didn't push my local branch, thanks!

PR #372, opened 2018-09-02
truncate() methods

tnagler:


include/vinecopulib/misc/triangular_array.hpp
```diff
@@ -150,6 +151,19 @@ void TriangularArray<T>::set_column(size_t column,
     mat_[column] = new_col;
 }
 
+//! truncates the trapezoid.
+//! @param trunc_lvl the truncation level.
+template<typename T>
+void TriangularArray<T>::truncate(size_t trunc_lvl)
+{
+    if (trunc_lvl < this->get_trunc_lvl()) {
```
tvatter: We should probably throw an error when this is not true to warn the user that the function isn't doing anything.

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -78,6 +78,7 @@ inline Vinecop::Vinecop(const std::vector<std::vector<Bicop>> &pair_copulas,
     check_pair_copulas_rvine_structure(pair_copulas);
 
     pair_copulas_ = pair_copulas;
+    vine_struct_.truncate(pair_copulas.size());
```
tvatter: Is it necessary?

include/vinecopulib/misc/triangular_array.hpp
tnagler: Not sure about that. Throwing an error is a very harsh way to tell a user that the function didn't do anything. The model will still be truncated at the requested level, just also a few levels before that.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Hm probably not necessary, but I think it's a good idea to keep things consistent. We don't really know how a user tries to obtain the truncation level. If we don't do this, we will give two different answers for `get_all_pair_copulas().size()` and `get_rvine_structure().get_trunc_lvl()`.

tvatter: OK

include/vinecopulib/misc/triangular_array.hpp
tvatter: Well, it's a typical warning situation, but there is no way (that I know of) to throw warnings in C++.

A few ideas there:

https://stackoverflow.com/questions/15449171/throw-runtime-warnings-in-c

tnagler: OK so the only sensible solution would be to send a message to `cout`/`clog`/`cerr`. Still not sure that is warranted though. Can you think of any situation where truncating at a higher-than-current level leads to unexpected behavior?

tvatter: I'm not sure, but it seems that the `if` warrants an `else` letting the user know that the function didn't do what it was expected to do. I would still favor the error throwing since the warning is not really an option... Btw, it's not really an error, but an exception, i.e. the program didn't do what it was intended to. In `Bicop::check_rotation` for instance, we're throwing an error while we could simply not do anything, but not doing anything while not letting the user know that nothing was done is usually bad practice.

tnagler: I think there's a difference between the two cases:
- If you rotate a Gaussian copula by 90 degrees, the sign of the parameter would change. So the user would rightfully expect the direction of dependence to change, which it doesn't. Hence, there is an error. (we could allow for that behavior btw)
- If you truncate a model at some level *k*, you'd expect only independence pair-copulas above level *k*. And this is what the user gets, even if the model is already *k-1* truncated.

tvatter:
Just one thing: maybe the function should be named `get_trunc_lvl` for consistency ?

tnagler:
I used the long name because we have `get/set_truncation_level()` in `FitControlsVinecop`. Noticed the inconsistency myself though.

tvatter:
Shouldn't we change this in `FitControlsVinecop` too then?

tnagler:
Yeah I'd prefer the shorter version too. Should we do [this](https://stackoverflow.com/questions/295120/c-mark-as-deprecated/21265197#21265197) to mark the old versions as deprecated before removing them altogether?

tvatter:
Oh that's a good idea.

PR #405, opened 2019-04-05
fix triangular array print method

tnagler:


include/vinecopulib/misc/triangular_array.hpp
```diff
@@ -207,17 +207,17 @@ std::string TriangularArray<T>::str() const
         }
         str << std::endl;
     }
-    return str.str();
+    return str.str().c_str();
```
tvatter: Why `c_str` ? Otherwise the PR is all good.

PR #407, opened 2019-04-05
refactoring for enhanced extensibility of Vinecop class

tnagler:


include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -332,9 +337,16 @@ inline FamilySelector::FamilySelector(const Eigen::MatrixXd &data,
 }
 
 inline StructureSelector::StructureSelector(const Eigen::MatrixXd &data,
-                                            const FitControlsVinecop &controls) 
+                                            const FitControlsVinecop &controls)
     : VinecopSelector(data, controls)
 {
+    n_ = data.rows();
```
tvatter: This stuff is already done in the `VinecopSelector` constructor.

include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -776,6 +791,23 @@ inline ptrdiff_t VinecopSelector::find_common_neighbor(size_t v0, size_t v1,
...
+// compute a fit id; can be used to re-use already fitted pair-copulas.
+//
+// @param edge.
+double VinecopSelector::compute_fit_id(const EdgeProperties& e)
+{
+    double id = 0.0;
+    if (controls_.needs_sparse_select()) {
+        // the formula is quite arbitrary, but sufficient for
+        // identifying situations where fits can be re-used
+        id = (e.pc_data.col(0) - 2 * e.pc_data.col(1)).sum();
```
tvatter: Sorry, I don't understand what should be achieved here.

tnagler: This was part of `select_pair_copulas()` before: https://github.com/vinecopulib/vinecopulib/pull/407/files#diff-42afd6af3367cd145f2fc7bb1d161b22L852

It is used in thresholded vines to use a fitted pair-copula form the last iteration if it hasn't changed. But using an already fitted pair-copula can be useful in a different context as well, that's why I made it a separate module.

include/vinecopulib/vinecop/implementation/tools_select.ipp
tnagler: true

include/vinecopulib/vinecop/implementation/tools_select.ipp
tvatter: Right, sorry. I still don't understand the formula though ^^

tnagler: There's not much too understand, because it's quite arbitrary ;) The only goal is to detect whether or not the input data has changed (with probability one). To also detect when the columns are exchanged, the formula needs to be asymmetric.

PR #412, opened 2019-04-26
Code health

tnagler:
mainly removing unused headers + a few simplifcations
- simulate_uniform has a qrng options, so we don't have to do the if switch all the time
- removed useless stuff in tools_stl

reduces the build times on my machine by about 10%

### DIFF
--- a/include/vinecopulib/bicop/elliptical.hpp
+++ b/include/vinecopulib/bicop/elliptical.hpp
@@ -9,6 +9,11 @@
 #include <vinecopulib/bicop/parametric.hpp>
 
 namespace vinecopulib {
+
+namespace constant {
+    constexpr double pi = 3.14159265358979323846;
+}
+
 //! @brief An abstract class for elliptical copula families
 //!
 //! This class is used in the implementation underlying the Bicop class.
--- a/include/vinecopulib/bicop/elliptical.hpp
+++ b/include/vinecopulib/bicop/elliptical.hpp
@@ -11,7 +11,7 @@
 namespace vinecopulib {
 
 namespace constant {
-    constexpr double pi = 3.14159265358979323846;
+    constexpr double pi = 3.141592653589793238462643383279502884;
 }
 
 //! @brief An abstract class for elliptical copula families
--- a/include/vinecopulib/bicop/implementation/bb6.ipp
+++ b/include/vinecopulib/bicop/implementation/bb6.ipp
@@ -6,6 +6,7 @@
 
 #include <vinecopulib/misc/tools_integration.hpp>
 #include <vinecopulib/misc/tools_stl.hpp>
+#include <vinecopulib/misc/tools_eigen.hpp>
 
 namespace vinecopulib {
 inline Bb6Bicop::Bb6Bicop()
@@ -128,6 +129,6 @@ inline double Bb6Bicop::parameters_to_tau(const Eigen::MatrixXd &parameters)
 
 inline Eigen::MatrixXd Bb6Bicop::tau_to_parameters(const double &tau)
 {
-    return vinecopulib::no_tau_to_parameters(tau);
+    return no_tau_to_parameters(tau);
 }
 }
--- a/include/vinecopulib/bicop/implementation/gaussian.ipp
+++ b/include/vinecopulib/bicop/implementation/gaussian.ipp
@@ -77,7 +77,7 @@ inline Eigen::VectorXd GaussianBicop::get_start_parameters(const double tau)
 inline Eigen::MatrixXd GaussianBicop::tau_to_parameters(const double &tau)
 {
     Eigen::VectorXd parameters = this->parameters_;
-    parameters(0) = sin(tau * constant::pi / 2);
+    parameters(0) = std::sin(tau * constant::pi / 2);
     return parameters;
 }
 }
--- a/include/vinecopulib/bicop/implementation/gaussian.ipp
+++ b/include/vinecopulib/bicop/implementation/gaussian.ipp
@@ -5,7 +5,6 @@
 // vinecopulib or https://vinecopulib.github.io/vinecopulib/.
 
 #include <vinecopulib/misc/tools_stats.hpp>
-#include <boost/math/constants/constants.hpp>
 
 namespace vinecopulib {
 inline GaussianBicop::GaussianBicop()
@@ -78,7 +77,7 @@ inline Eigen::VectorXd GaussianBicop::get_start_parameters(const double tau)
 inline Eigen::MatrixXd GaussianBicop::tau_to_parameters(const double &tau)
 {
     Eigen::VectorXd parameters = this->parameters_;
-    parameters(0) = sin(tau * boost::math::constants::pi<double>() / 2);
+    parameters(0) = sin(tau * constant::pi / 2);
     return parameters;
 }
 }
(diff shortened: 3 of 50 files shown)


include/vinecopulib/bicop/elliptical.hpp
```diff
@@ -9,6 +9,11 @@
 #include <vinecopulib/bicop/parametric.hpp>
 
 namespace vinecopulib {
+
+namespace constant {
+    constexpr double pi = 3.14159265358979323846;
```
tvatter: Is there a good reason for removing the ifndef? The M_PI is a standard I think...

include/vinecopulib/bicop/implementation/bb6.ipp
```diff
@@ -6,6 +6,7 @@
 
 #include <vinecopulib/misc/tools_integration.hpp>
 #include <vinecopulib/misc/tools_stl.hpp>
```
tvatter: This header and the previous one become unnecessary no?

include/vinecopulib/bicop/implementation/gaussian.ipp
```diff
@@ -78,7 +77,7 @@ inline Eigen::VectorXd GaussianBicop::get_start_parameters(const double tau)
 inline Eigen::MatrixXd GaussianBicop::tau_to_parameters(const double &tau)
 {
     Eigen::VectorXd parameters = this->parameters_;
-    parameters(0) = sin(tau * boost::math::constants::pi<double>() / 2);
+    parameters(0) = sin(tau * constant::pi / 2);
```
tvatter: It's funny that cran does not complain about std::sin here no?

include/vinecopulib/bicop/implementation/kernel.ipp
```diff
@@ -15,7 +16,7 @@ inline KernelBicop::KernelBicop()
     Eigen::VectorXd grid_points(m);
     for (size_t i = 0; i < m; ++i)
         grid_points(i) = -3.25 + i * (6.5 / static_cast<double>(m - 1));
-    interp_grid_ = tools_interpolation::InterpolationGrid(
+    interp_grid_ = std::make_shared<tools_interpolation::InterpolationGrid>(
```
tvatter: What is the advantage of doing that?

include/vinecopulib/misc/tools_eigen.hpp
```diff
@@ -21,7 +21,11 @@ template<typename T>
 Eigen::MatrixXd unaryExpr_or_nan(const Eigen::MatrixXd &x, const T &func)
 {
     return x.unaryExpr([&func](const double &y) {
-        return tools_stl::unaryFunc_or_nan(func, y);
```
tvatter: Why is your thing better?

tvatter: (and same question for the binary)

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -727,17 +730,9 @@ Vinecop::cdf(const Eigen::MatrixXd &u, const size_t N,
     check_data_dim(u);
 
     // Simulate N quasi-random numbers from the vine model
-    Eigen::MatrixXd U(N, d_);
-    if (d_ > 300) {
-        U = tools_stats::sobol(N, d_, seeds);
-    } else {
-        U = tools_stats::ghalton(N, d_, seeds);
-    }
+    auto U = tools_stats::simulate_uniform(N, d_, true, seeds);
```
tvatter: Maybe here it should be switched from U to u too no?

tvatter: Or something else since u is already taken.

include/vinecopulib/bicop/elliptical.hpp
tnagler: yeah from C ;) constexpr is modern C++

include/vinecopulib/bicop/implementation/gaussian.ipp
tnagler: CRAN is totally inconsistent. I have no clue what it explains about and what not ^^

include/vinecopulib/bicop/implementation/kernel.ipp
tnagler: interpolation_grid.hpp  doesn't need to be included. Hence it is no longer in every file that includes kernel.hpp (which is every file that includes bicop.hpp)

include/vinecopulib/misc/tools_eigen.hpp
tnagler: tools_stl header no leader needs to be included in tools_eigen. The stuff in tools_stl was only used within the tools_eigen header, so why not have the code in tools_eigen directly.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: u_sim?

include/vinecopulib/bicop/elliptical.hpp
tvatter: My issue is not with the constexpr, it's with using our own definition. How about using stuff from [boost](https://www.boost.org/doc/libs/1_37_0/libs/math/doc/sf_and_dist/html/math_toolkit/toolkit/internals1/constants.html) ?

include/vinecopulib/bicop/implementation/gaussian.ipp
tvatter: True, but in any case, should'nt we use std::sin?

include/vinecopulib/bicop/implementation/kernel.ipp
tvatter: That's cool :)

include/vinecopulib/misc/tools_eigen.hpp
tvatter: OK

include/vinecopulib/vinecop/implementation/class.ipp
tvatter: Sure

include/vinecopulib/misc/tools_bobyqa.hpp
```diff
@@ -16,8 +16,8 @@ namespace vinecopulib {
 
 namespace tools_bobyqa {
 
-constexpr double sqrt_2 = std::sqrt(2.0);
-constexpr double sqrt_0_5 = std::sqrt(0.5);
+constexpr double sqrt_2 = 1.41421356237309504880168872420969807;
```
tvatter: Same as before, I think that we should use [boost](https://www.boost.org/doc/libs/1_37_0/libs/math/doc/sf_and_dist/html/math_toolkit/toolkit/internals1/constants.html)

include/vinecopulib/bicop/elliptical.hpp
tnagler: Using our own def is just more leightweight, but we can use boost if you prefer (I will add more digits in the meanwhile).

include/vinecopulib/bicop/implementation/gaussian.ipp
tnagler: will do

PR #414, opened 2019-04-30
fix adjustment for Brent bounds

tnagler:
Actual changes are L125-130. Am I stupid or was this completely wrong the whole time?


tvatter:
Wow, how did you notice this?

tnagler:
Was coincidentally looking at the code while fixing a bug in the t-vine branch.

tvatter:
Good catch, I'm surprised that this went unnoticed...

PR #419, opened 2019-05-02
Enhance extensibility

tnagler:
- getters for parameter bounds of `Bicop` objects
- private members of `RVineStructure` become protected, so that derived classes can access them
- some missing `inline`s were added (don't we test the shared lib?)


include/vinecopulib/bicop/implementation/kernel.ipp
```diff
@@ -88,6 +88,16 @@ inline Eigen::MatrixXd KernelBicop::get_parameters() const
...
 }
 
+inline Eigen::MatrixXd KernelBicop::get_parameters_lower_bounds() const
+{
+    return Eigen::MatrixXd::Constant(30, 30, 0.0);
+}
+
+inline Eigen::MatrixXd KernelBicop::get_parameters_upper_bounds() const
+{
+    return Eigen::MatrixXd::Constant(30, 30, 1e4);
```
tvatter: Really?

PR #428, opened 2019-07-10
Switch to quadrature for integration

tvatter:


include/vinecopulib/misc/implementation/tools_integration.ipp
```diff
@@ -0,0 +1,67 @@
...
+    case 30:
+      output = tools_legendre::nodes_weights_30;
+      break;
+    case 40:
+      output = tools_legendre::nodes_weights_40;
+      break;
+    case 50:
+      output = tools_legendre::nodes_weights_50;
+      break;
+    default:
```
tnagler: I assume that we will always just use the default `n` (currently 20). I'd vote to keep the code base lean by removing the option altogether, possibly changing the default to something a bit larger. Did you do some accuracy checks?

tnagler: Started a preliminary implementation of conditional distributions. Seems like we need **at least** `n = 50`. Relative error was still around 5-10% :(

tvatter: Yeah, that was at that point that I stopped: I could not get the errors at a acceptably low level :(

tnagler:
OK, how about we put this on hold and reconsider for the next release?

PR #446, opened 2019-07-23
Ensure consistency in manually created TLL Bicop (fixes #440)

### LINKED ISSUE #440: Loglik of TLL family
For some reason, there's a discrepancy in the "fitted" loglik and the one that we compute from a fitted model, see R code below.
```
library(rvinecopulib)
library(VineCopula)
data(daxreturns)
u1 <- daxreturns[,1]
u2 <- daxreturns[,2]

fit <- bicop(cbind(u1,u2), family_set = "tll", nonpar_method = "constant")
fit$loglik
#[1] 297.3605
sum(log(dbicop(cbind(u1, u2), fit)))
#[1] 299.9817
```

tnagler:
Ensures that `cop == cop_new`, where
``` cpp

auto cop = Bicop(u, FitControlsBicop({BicopFamily::tll}});
auto cop_new = Bicop(BicopFamily::tll, 0, cop.get_parameters());
```
This is what we do when converting R and C++ objects.

### DIFF
--- a/include/vinecopulib/bicop/implementation/kernel.ipp
+++ b/include/vinecopulib/bicop/implementation/kernel.ipp
@@ -16,8 +16,14 @@ inline KernelBicop::KernelBicop()
   Eigen::VectorXd grid_points(m);
   for (size_t i = 0; i < m; ++i)
     grid_points(i) = -3.25 + i * (6.5 / static_cast<double>(m - 1));
+  grid_points = tools_stats::pnorm(grid_points);
+  
+  // move boundary points to 0/1, so we don't have to extrapolate
+  grid_points(0) = 0.0;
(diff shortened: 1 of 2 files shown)


tvatter:
Can you just explain why this solves the problem?

tnagler:
1. After fitting a tll, the boundary points of the grid are moved to 0 and 1. This was not the case for a default constructed tll Hence, the grid for interpolation was different for `cop` and `cop_new`. This is fixed by (8969689).

2. When setting the parameters of a default constructed tll, the values were normalized to approximately uniform margins in 3 more iterations. Hence, also the values on the interpolation grids were different  for `cop` and `cop_new`. This is fixed by (f0950d6).

Issue #438, opened 2019-07-19
Memory efficiency in vine algorithms


tnagler:
These blocks are unnecessary 
```
  if (trunc_lvl > 0) {
    order = vine_struct_.get_order();
    no_array = vine_struct_.get_struct_array(true);
    min_array = vine_struct_.get_min_array();
    needed_hfunc1 = vine_struct_.get_needed_hfunc1();
    needed_hfunc2 = vine_struct_.get_needed_hfunc2();
  }
```
We just need to export `needed_hfunc1/2(i, j)` methods.

tnagler:
+ the redundant copies introduced by discrete variables (#454).

tnagler:
See #460.

PR #454, opened 2019-09-26
Discrete variables

tnagler:
So here's a first draft for discussion.

### API choices
**Reminder**
To compute pdfs etc for discrete variables we need not only $F_X(X)$ but also $F_X(X^-),$ where $X^- = X - 1$ for integer valued variables. Since we work on the copula scale, we thus need two pseudo-variables instead of one. 

**Passing data to functions**
I didn't want to introduce extra arguments to all our functions, so my idea is that the user supplies a `n x 2d` matrix as soon as discrete variables are present. The first `n x d` block is for $F_X(X)$ and the second for $F_X(X^-)$.

**How to know which variables are discrete**
Currently, the user tells us which variables are discrete by, e.g., `model.set_var_types({"c", "d"})`. Another option would be to supply a vector of numbers indicating the discrete columns. 

There is also potential for auto-detection by checking for equality between columns (by checking which columns are identical, we would know which variables are discrete). However, it's impossible to exclude errors I think.

### TODOS
- [ ] Extensive unit tests
- [ ] Adapt documentation
- [ ] Clean code


tnagler:
Alright, I added some unit tests, serialization, and a bit of documentation.
Along the way, I fixed two minor things:
- I noticed that `hinv` was always returning 0 for a Clayton with parameter 0, so I set the default and lower bound to 1e-10.
- I added `loglik`, `threshold`, and `nobs` to the serialized objects, so no information about the fit gets lost.

tvatter:
For instance, the signature of `Bicop::pdf` would go from:

```
Eigen::VectorXd pdf(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u) const;
```

To:

```
Eigen::VectorXd pdf(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u,
                                 const std::vector<bool> discrete = {false, false},
                                 const Eigen::MatrixXd& u2 = Eigen::MatrixXd()) const;
```

tnagler:
More reasoning for the current choice:
1. No breaking changes for continuous data either.
2. I don't think that the increased memory is a problem. If the data doesn't fit in memory twice, I doubt that you can use the library anyway (multiples of the original data's size is allocated when fitting a vine). 
3.  Currently there is no additional argument at all, just one additional member function.
4. Indeed, the API change for `Bicop::pdf()` and similar functions wouldn't be an issue. However, there are much more problematic instances. We would need additional overloads for at least:

   - one `Bicop()` constructor.
   - `Bicop::select()` and `Bicop::fit()`
   - `Bicop::mbic()`
   - four (!) `Vinecop()` constructors
   - all of `Vinecop's` pdf/cdf/rosenblatt/aic etc (they have a `num_threads` argument, moving it would be a breaking change)

Especially regarding 4, I'm a bit afraid about maintainability.

tvatter:
How is that not a breaking change for continuous data?

I didn't think of 4, that's an excellent point.

Something that would be "in-between" would be a (d+k)n matrix as input...

tnagler:
For continuous variables only, you can just supply n x d as before. Currently, you're even required to make it n x d (not n x 2d), but we could relax that.

n x (d+k) would be an option. I personally think that's less convenient (but that's debatable since I don't have any experience in actual use cases). 

Generally, I think convenience is the most important factor. Memory requirements of fitting a vine are around n x 4d + pair copulas. Whether that goes to n x 5d or n x (4d + k) shouldn't matter much.

tnagler:
Good point, I think n x (d + k) it is then. I should also mention that everything in here is a C++ only discussion. We could easily have non-breaking additional arguments in R/Python that can be supplied by name.

include/vinecopulib/bicop/abstract.hpp
```diff
@@ -39,6 +39,8 @@ class AbstractBicop
 
   void set_loglik(const double loglik = NAN);
 
+  void set_var_types(const std::vector<std::string>& var_types);
```
tvatter: Also a getter maybe?

include/vinecopulib/bicop/abstract.hpp
```diff
@@ -47,52 +49,60 @@ class AbstractBicop
...
 
   virtual double parameters_to_tau(const Eigen::MatrixXd& parameters) = 0;
 
-  Eigen::VectorXd pdf(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u);
+  virtual void flip() = 0;
 
-  double loglik(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u,
-                const Eigen::VectorXd weights = Eigen::VectorXd());
+  // following are virtual so they can be overriden by KernelBicop
+  virtual Eigen::VectorXd pdf(const Eigen::MatrixXd& u);
```
tvatter: Can you explain why you changed some by adding the ` = 0` and for others you removed it?

include/vinecopulib/bicop/class.hpp
```diff
@@ -119,24 +115,40 @@ class Bicop
 
   Eigen::MatrixXd get_parameters_upper_bounds() const;
 
+  Bicop as_continuous() const;
+
 private:
-  Eigen::Matrix<double, Eigen::Dynamic, 2> cut_and_rotate(
-    const Eigen::Matrix<double, Eigen::Dynamic, 2>& u) const;
+  Eigen::MatrixXd format_data(const Eigen::MatrixXd& u) const;
```
tvatter: Why the renaming?

include/vinecopulib/bicop/class.hpp
```diff
@@ -119,24 +115,40 @@ class Bicop
...
+  void check_data_dim(const Eigen::MatrixXd& u) const;
+
+  void flip_var_types();
+
   void check_weights_size(const Eigen::VectorXd& weights,
                           const Eigen::MatrixXd& data) const;
 
   void check_fitted() const;
 
+  int get_n_discrete() const;
```
tvatter: Following the trace of where this function was used made me think: currently, we use `format_data` everytime we try to recompute the log-likelihood in the fit. It's really super inefficient, when has this changed? At some point, we had a way to compute a "rough" log-likelihood that was not re-formatting the data at every function call...

include/vinecopulib/bicop/implementation/abstract.ipp
```diff
@@ -20,6 +21,16 @@
 #include <vinecopulib/bicop/tll.hpp>
 #include <vinecopulib/misc/tools_eigen.hpp>
 
+
+
+
+
+#include <iostream>
```
tvatter: Debug leftover or actually needed?

include/vinecopulib/bicop/bb8.hpp
```diff
@@ -33,8 +33,8 @@ class Bb8Bicop : public ArchimedeanBicop
 
   double generator_derivative2(const double& u);
 
-  // pdf
-  Eigen::VectorXd pdf_raw(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u);
+    // pdf
+    Eigen::VectorXd pdf_raw(const Eigen::MatrixXd &u);
```
tvatter: pdf_c_c and along with the others?

include/vinecopulib/bicop/implementation/abstract.ipp
```diff
@@ -118,25 +129,116 @@ AbstractBicop::set_loglik(const double loglik)
...
     return std::min(DBL_MAX, std::max(x, DBL_MIN));
   };
-  return tools_eigen::unaryExpr_or_nan(pdf_raw(u), trim);
+  return tools_eigen::unaryExpr_or_nan(pdf, trim);
+}
+
+inline Eigen::VectorXd
+AbstractBicop::pdf_c_d(const Eigen::MatrixXd& u)
+{
+  auto umax = u.leftCols(2);
```
tvatter: Do we really have to do all this copying? Same question for all the functions below...

include/vinecopulib/bicop/implementation/abstract.ipp
```diff
@@ -118,25 +129,116 @@ AbstractBicop::set_loglik(const double loglik)
...
+  auto umin = u.rightCols(2);
+  if (var_types_[0] != "c") {
+    return (hfunc2_raw(umax) - hfunc2_raw(umin)).array().abs();
```
tvatter: Stupid question, but why the abs()?

include/vinecopulib/bicop/implementation/bb6.ipp
```diff
@@ -7,6 +7,7 @@
 #include <vinecopulib/misc/tools_eigen.hpp>
 #include <vinecopulib/misc/tools_integration.hpp>
 #include <vinecopulib/misc/tools_stl.hpp>
+#include <vinecopulib/misc/tools_eigen.hpp>
```
tvatter: Why is that needed? Same question below...

include/vinecopulib/bicop/implementation/bb8.ipp
```diff
@@ -6,6 +6,7 @@
 
 #include <vinecopulib/misc/tools_eigen.hpp>
 #include <vinecopulib/misc/tools_integration.hpp>
+#include <vinecopulib/misc/tools_eigen.hpp>
```
tvatter: Why is that needed?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -54,7 +54,15 @@ inline Bicop::Bicop(const boost::property_tree::ptree input)
           input.get<int>("rotation"),
           tools_serialization::ptree_to_matrix<double>(
             input.get_child("parameters")))
-{}
+{
+  var_types_ = tools_serialization::ptree_to_vector<std::string>(
+    input.get_child("var_types"));
+  // try block for backwards compatibility
```
tvatter: If you really want backwards compatibility, shouldn't you also include the previous 2 lines in the try block?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -54,7 +54,15 @@ inline Bicop::Bicop(const boost::property_tree::ptree input)
...
+  try {
+    nobs_ = input.get<size_t>("nobs_");
+    bicop_->set_loglik(input.get<double>("loglik"));
+  } catch (...) {}
```
tvatter: Is the `catch(...) {}` really useful here?

include/vinecopulib/bicop/class.hpp
```diff
@@ -119,24 +115,40 @@ class Bicop
...
+
   double compute_mbic_penalty(const size_t nobs, const double psi0) const;
 
   BicopPtr get_bicop() const;
 
   BicopPtr bicop_;
   int rotation_;
-  size_t nobs_;
+  size_t nobs_{0};
+  mutable std::vector<std::string> var_types_{ "c", "c" };
```
tvatter: Why are you adding this one a second time? Isn't it in abstract already?

Unrelated, but now that I think about it, why is `loglik` in abstract and `nobs` here?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -97,36 +110,35 @@ Bicop::to_json(const std::string filename) const
...
+//! @param data an \f$ n \times (2 + k) \f$ matrix of observations contained in
+//!   \f$(0, 1)^2 \f$, where \f$ k \f$ is the number of discrete variables.
 //! @return The copula distribution evaluated at \c u.
 inline Eigen::VectorXd
-Bicop::cdf(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u) const
+Bicop::cdf(const Eigen::MatrixXd& u) const
 {
-  tools_eigen::check_if_in_unit_cube(u);
-  Eigen::VectorXd p = bicop_->cdf(cut_and_rotate(u));
+  check_data(u);
```
tvatter: Why do we need to pass e.g. a matrix with 4 columns if need only 2 of those?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -252,15 +298,16 @@ Bicop::simulate(const size_t& n,
...
+//! @param data an \f$ n \times (2 + k) \f$ matrix of observations contained in
+//!   \f$(0, 1)^2 \f$, where \f$ k \f$ is the number of discrete variables.
 inline double
-Bicop::loglik(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u) const
+Bicop::loglik(const Eigen::MatrixXd& u) const
 {
   if (u.rows() < 1) {
     return get_loglik();
   } else {
     tools_eigen::check_if_in_unit_cube(u);
```
tvatter: Not `check_data`?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -462,17 +511,80 @@ inline void
...
+    std::stringstream msg;
+    msg << "data has wrong number of columns; "
+        << "expected: " << n_cols_exp << " or 4, actual: " << n_cols 
+        << " (model contains ";
+    if (n_cols_exp == 2) {
+      msg << "no ";
+    } else {
+      msg << get_n_discrete() << " ";
+    }
+    msg << "discrete variables)." << std::endl;
```
tvatter: I'm going to be annoying by pointing out that it could print `(model contains 1 discrete variables)`, which is grammatically incorrect :) JK, I don't care.

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -462,17 +511,80 @@ inline void
...
+//!   e.g., `{"c", "d"}` means first varible continuous, second discrete.
+inline void
+Bicop::set_var_types(const std::vector<std::string>& var_types)
+{
+  if (var_types.size() != 2) {
+    throw std::runtime_error("var_types must have size two.");
+  }
+  for (auto t : var_types) {
+    if (!tools_stl::is_member(t, { "c", "d" })) {
+      throw std::runtime_error("var type must be either 'c' or 'd'.");
```
tvatter: The components of var_types must be either?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -536,7 +649,18 @@ Bicop::get_bicop() const
   return bicop_;
 }
 
-//! @brief fits a bivariate copula (with fixed family) to data.
+inline Bicop
```
tvatter: no documentation for this one?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -657,19 +804,46 @@ Bicop::select(const Eigen::Matrix<double, Eigen::Dynamic, 2>& data,
...
+//! removes superfluous columns for continuous variables. 
+//! (continuous models only require two columns, discrete models always four)
+inline Eigen::MatrixXd
+Bicop::format_data(const Eigen::MatrixXd& u) const
 {
-  Eigen::Matrix<double, Eigen::Dynamic, 2> u_new(u.rows(), 2);
+  int n_disc = get_n_discrete();
+  if (n_disc == 0) {
+    return u.leftCols(2);
+  } else if ((get_n_discrete() != 1) | (u.cols() == 4)) {
```
tvatter: n_disc

include/vinecopulib/bicop/class.hpp
tvatter: size_t ?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -657,19 +804,46 @@ Bicop::select(const Eigen::Matrix<double, Eigen::Dynamic, 2>& data,
...
+    return u;
+  }
+  Eigen::MatrixXd u_new(u.rows(), 4);
+  u_new.leftCols(2) = u.leftCols(2);
+  int disc_col = (var_types_[1] == "d");
```
tvatter: size_t ?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -657,19 +804,46 @@ Bicop::select(const Eigen::Matrix<double, Eigen::Dynamic, 2>& data,
...
+  u_new.col(2 + disc_col) = u.col(2);
+  u_new.col(2 + cont_col) = u.col(cont_col);
+  return u_new;
+}
 
+//! clisp the data to the interval [1e-10, 1 - 1e-10] for numerical stability.
+inline Eigen::MatrixXd
+Bicop::clip_data(const Eigen::MatrixXd& u) const
+{
+  auto clip = [] (const double& x) {
```
tvatter: So, I've seen the trim function written a few times, always followed by `tools_eigen::unaryExpr_or_nan`  already above, and now this clip function... Shouldn't we add something in tools_eigen?

Like
```
Eigen::VectorXd trim(const Eigen::VectorXd& x, const down = 0, const up = o) const {
  auto trim_one = [down, up] (const double& x) {
      return std::min(std::max(x, down), 1 - up);
  };
  return tools_eigen::unaryExpr_or_nan(x, trim_one);
}
```

include/vinecopulib/bicop/implementation/kernel.ipp
```diff
@@ -6,6 +6,7 @@
 
 #include <vinecopulib/misc/tools_interpolation.hpp>
 #include <vinecopulib/misc/tools_stats.hpp>
+#include <vinecopulib/misc/tools_interpolation.hpp>
```
tvatter: Why?

include/vinecopulib/bicop/implementation/kernel.ipp
```diff
@@ -16,50 +17,80 @@ inline KernelBicop::KernelBicop()
...
   );
 }
 
 inline Eigen::VectorXd
-KernelBicop::pdf_raw(const Eigen::Matrix<double, Eigen::Dynamic, 2>& u)
+KernelBicop::pdf_raw(const Eigen::MatrixXd& u)
+{
+  auto pdf = interp_grid_->interpolate(u);
+  auto trunc = [] (const double& p) {
+      return std::fmax(p, 1e-20);
```
tvatter: Why not `std::max` ?

include/vinecopulib/bicop/implementation/kernel.ipp
```diff
@@ -16,50 +17,80 @@ inline KernelBicop::KernelBicop()
   Eigen::VectorXd grid_points(m);
   for (size_t i = 0; i < m; ++i)
     grid_points(i) = -3.25 + i * (6.5 / static_cast<double>(m - 1));
-  grid_points = tools_stats::pnorm(grid_points);
-  
-  // move boundary points to 0/1, so we don't have to extrapolate
-  grid_points(0) = 0.0;
```
tvatter: Why don't we need that anymore?

include/vinecopulib/bicop/implementation/kernel.ipp
```diff
@@ -16,50 +17,80 @@ inline KernelBicop::KernelBicop()
...
+  };
+  return tools_eigen::unaryExpr_or_nan(pdf, trunc);
+}
+
+inline Eigen::VectorXd
+KernelBicop::pdf(const Eigen::MatrixXd& u)
 {
-  return interp_grid_->interpolate(u);
+  if (u.cols() == 4) {
+    // evaluate jittered density at mid rank for stability
```
tvatter: Can you explain? Why do we treat the kernel differently from the others?

tvatter: (same question for the hfuncs/hinvs)

include/vinecopulib/bicop/implementation/tll.ipp
```diff
@@ -228,8 +227,11 @@ TllBicop::fit(const Eigen::Matrix<double, Eigen::Dynamic, 2>& data,
   auto grid_2d = tools_eigen::expand_grid(grid_points);
 
   // transform evaluation grid and data by inverse Gaussian cdf
-  Eigen::Matrix<double, Eigen::Dynamic, 2> z = tools_stats::qnorm(grid_2d);
-  Eigen::Matrix<double, Eigen::Dynamic, 2> z_data = tools_stats::qnorm(data);
+  Eigen::MatrixXd z = tools_stats::qnorm(grid_2d);
+
+  // use jittering in case observations are discrete
+  auto psobs = tools_stats::to_pseudo_obs(data.leftCols(2), "random");
```
tvatter: Shouldn't this be documented? And maybe we should also have a way to pass a seed so that the fits are reproducible no?

include/vinecopulib/bicop/implementation/tll.ipp
```diff
@@ -251,14 +253,21 @@ TllBicop::fit(const Eigen::Matrix<double, Eigen::Dynamic, 2>& data,
...
   // don't normalize margins of the EDF! (norm_times = 0)
   auto infl_grid = InterpolationGrid(grid_points, infl, 0);
-  npars_ = infl_grid.interpolate(data).sum();
-  set_loglik(loglik(data, weights));
+  if ((var_types_[0] == "d") | (var_types_[1] == "d")) {
+      // for discrete, use mid ranks to compute EDF and log-likelihood
+      // (this is closer to "observations" than jittered or "upper" pseudo data)
+      psobs = 0.5 * (data.leftCols(2) + data.rightCols(2)).array();
+      npars_ = tools_eigen::unique(infl_grid.interpolate(psobs)).sum();
+      npars_ = std::fmax(npars_, 1.0);
```
tvatter: Why fmax rather than max?

include/vinecopulib/bicop/implementation/tll.ipp
```diff
@@ -251,14 +253,21 @@ TllBicop::fit(const Eigen::Matrix<double, Eigen::Dynamic, 2>& data,
...
   infl = Eigen::Map<Eigen::MatrixXd>(infl_vec.data(), m, m).transpose();
   // don't normalize margins of the EDF! (norm_times = 0)
   auto infl_grid = InterpolationGrid(grid_points, infl, 0);
-  npars_ = infl_grid.interpolate(data).sum();
-  set_loglik(loglik(data, weights));
+  if ((var_types_[0] == "d") | (var_types_[1] == "d")) {
+      // for discrete, use mid ranks to compute EDF and log-likelihood
+      // (this is closer to "observations" than jittered or "upper" pseudo data)
+      psobs = 0.5 * (data.leftCols(2) + data.rightCols(2)).array();
+      npars_ = tools_eigen::unique(infl_grid.interpolate(psobs)).sum();
```
tvatter: Are we preventing the NaNs somewhow here?

include/vinecopulib/bicop/implementation/tll.ipp
```diff
@@ -251,14 +253,21 @@ TllBicop::fit(const Eigen::Matrix<double, Eigen::Dynamic, 2>& data,
...
+      npars_ = std::fmax(npars_, 1.0);
+  } else {
+    npars_ = std::fmax(infl_grid.interpolate(data).sum(), 1.0);
+  }
+  set_loglik(pdf(data).array().log().sum());
```
tvatter: Same question about fmax as well as the NaNs.

include/vinecopulib/misc/implementation/tools_eigen.ipp
```diff
@@ -75,13 +75,23 @@ check_if_in_unit_cube(const Eigen::MatrixXd& u)
...
 {
   u.col(0).swap(u.col(1));
   return u;
 }
 
+inline Eigen::VectorXd
+unique(const Eigen::VectorXd& x)
+{
+  std::vector<double> v(x.data(), x.data() + x.size());
+  std::sort(v.begin(), v.end()); // 1 1 2 2 3 3 3 4 4 5 5 6 7
```
tvatter: Why do you need to sort and then erase the elements from `v`? It seems like a lot of useless computations no? Or you need them sorted on the other side?

The reason that I ask is because using a hash-map would reduce to o(n) average time: https://www.geeksforgeeks.org/print-distinct-elements-given-integer-array/

include/vinecopulib/vinecop/class.hpp
```diff
@@ -81,11 +81,11 @@ class Vinecop
   void to_json(const std::string filename) const;
 
   // Methods modifying structure and/or families and parameters
-  void select_all(const Eigen::MatrixXd& data,
```
tvatter: Why?

include/vinecopulib/misc/triangular_array.hpp
```diff
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <iostream>
+#include <vector>
```
tvatter: Why?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -19,29 +19,21 @@ namespace vinecopulib {
...
 //! @brief creates a vine copula with structure specified by an RVineStructure
 //! object; all pair-copulas are set to independence.
 //! @param structure an RVineStructure object representing the structure of
 //! the vine.
 inline Vinecop::Vinecop(const RVineStructure& structure)
-{
-  d_ = structure.get_dim();
-  vine_struct_ = structure;
+  : d_(structure.get_dim())
+  , vine_struct_(structure)
```
tvatter: the , have to be on the next line ?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -19,29 +19,21 @@ namespace vinecopulib {
...
-  loglik_ = NAN;
-}
+  : Vinecop(RVineStructure(tools_stl::seq_int(1, d), static_cast<size_t>(0)))
+{}
 
 //! @brief creates a vine copula with structure specified by an RVineStructure
 //! object; all pair-copulas are set to independence.
 //! @param structure an RVineStructure object representing the structure of
 //! the vine.
 inline Vinecop::Vinecop(const RVineStructure& structure)
```
tvatter: I noticed that you change the next `structure` for `vine_struct` but not this one. Why?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -72,20 +64,14 @@ inline Vinecop::Vinecop(const std::vector<size_t>& order,
...
-
-  d_ = structure.get_dim();
-  vine_struct_ = structure;
-
   check_pair_copulas_rvine_structure(pair_copulas);
-
   pair_copulas_ = pair_copulas;
   vine_struct_.truncate(pair_copulas.size());
-  threshold_ = 0.0;
-  loglik_ = NAN;
```
tvatter: Should we maybe set the loglik if the pair copulas have already a loglik?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -146,9 +132,15 @@ inline Vinecop::Vinecop(const boost::property_tree::ptree input,
       pair_copulas_[tree][edge] = Bicop(pc_node);
     }
   }
-
-  threshold_ = 0;
-  loglik_ = NAN;
+  var_types_ = tools_serialization::ptree_to_vector<std::string>(
+    input.get_child("var_types"));
+
+  // try block for backwards compatibility
```
tvatter: Same comment as for bicop regarding having var_types in the try block (and also about the (...) {}).

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -168,18 +160,17 @@ inline Vinecop::Vinecop(const std::string filename, const bool check)
...
-                        const RVineStructure& structure,
+                        const RVineStructure& vine_struct,
                         FitControlsVinecop controls)
+  : Vinecop(vine_struct)
 {
-  d_ = data.cols();
   nobs_ = data.rows();
   check_enough_data(data);
-  if (d_ != structure.get_dim()) {
+  if (static_cast<size_t>(data.cols()) != vine_struct_.get_dim()) {
```
tvatter: Why don't you use `d_` since it is set before?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -296,46 +292,66 @@ Vinecop::make_pair_copula_store(const size_t d, const size_t trunc_lvl)
...
+//! left-sided limit of the cdf. For continuous variables the left limit and the
+//! cdf itself coincide. For, e.g., an integer-valued variable, it holds \f$
+//! F_Y(Y^-) = F_Y(Y - 1) \f$. Continuous variables in the second block can
+//! be omitted.
+//!
+//! @param data \f$ n \times (d + k) \f$ or \f$ n \times 2d \f$ matrix of
+//!   observations, where \f$ k \f$ is the number of discrete variables.
 //! @param controls the controls to the algorithm (see FitControlsVinecop).
 inline void
-Vinecop::select_all(const Eigen::MatrixXd& data,
```
tvatter: Why?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -630,20 +645,80 @@ Vinecop::get_threshold() const
...
+Vinecop::set_var_types(const std::vector<std::string>& var_types)
+{
+  std::stringstream msg;
+  if (var_types.size() > d_) {
+    msg << "more var_types (" << var_types.size() << ")"
+        << "than variables (" << d_ << ")-" << std::endl;
+  }
+  for (auto t : var_types) {
+    if (!tools_stl::is_member(t, { "c", "d" })) {
+      msg << "variable type must be 'c' or 'd'." << std::endl;
```
tvatter: So the error message could be a list of `"variable type must be 'c' or 'd'."` statements one after the other? Shouldn't we write only once? Or let the user know which component is problematic?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -630,20 +645,80 @@ Vinecop::get_threshold() const
...
+//! @param var_types a vector of size two specifying the types of the variables,
+//!   e.g., `{"c", "d"}` means first varible continuous, second discrete.
+inline void
+Vinecop::set_var_types(const std::vector<std::string>& var_types)
+{
+  std::stringstream msg;
+  if (var_types.size() > d_) {
+    msg << "more var_types (" << var_types.size() << ")"
+        << "than variables (" << d_ << ")-" << std::endl;
+  }
```
tvatter: I would throw the error here already, and then throw another one if the type is incorrect, otherwise the message could look very weird no?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -630,20 +645,80 @@ Vinecop::get_threshold() const
...
+  if (!msg.str().empty()) {
+    throw std::runtime_error(msg.str());
+  }
+  var_types_ = var_types;
+  if (pair_copulas_.size() == 0) {
+    return;
+  }
+
+  // set new var_types for all pair-copulas
+  std::vector<std::string> no_types(d_), pair_types(2);
```
tvatter: Those variable names with `no` are always annoying ^^ I'm pretty sure that someone that don't know about natural order would have a hard time here. But I don't really care.

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -657,46 +732,72 @@ Vinecop::pdf(const Eigen::MatrixXd& u, const size_t num_threads) const
...
 
     // fill first row of hfunc2 matrix with evaluation points;
     // points have to be reordered to correspond to natural order
-    for (size_t j = 0; j < d; ++j)
+
+    for (size_t j = 0; j < d; ++j) {
       hfunc2.col(j) = u.block(b.begin, order[j] - 1, b.size, 1);
+      if (var_types_[order[j] - 1] == "d") {
+        hfunc2_sub.col(j) =
+          u.block(b.begin, d_ + disc_cols[order[j] - 1], b.size, 1);
```
tvatter: Sorry, I don't get it, why are we using data at u - 1 for the discrete variables here?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -657,46 +732,72 @@ Vinecop::pdf(const Eigen::MatrixXd& u, const size_t num_threads) const
...
 
     for (size_t tree = 0; tree < trunc_lvl; ++tree) {
       tools_interface::check_user_interrupt(n * d > 1e5);
       for (size_t edge = 0; edge < d - tree - 1; ++edge) {
         tools_interface::check_user_interrupt(edge % 100 == 0);
         // extract evaluation point from hfunction matrices (have been
         // computed in previous tree level)
-        size_t m = min_array(tree, edge);
         u_e.col(0) = hfunc2.col(edge);
+        u_e.col(2) = hfunc2_sub.col(edge);
```
tvatter: So we do this even when both variables are continuous?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -657,46 +732,72 @@ Vinecop::pdf(const Eigen::MatrixXd& u, const size_t num_threads) const
...
+        size_t m = min_array(tree, edge);
         if (m == no_array(tree, edge)) {
           u_e.col(1) = hfunc2.col(m - 1);
+          u_e.col(3) = hfunc2_sub.col(m - 1);
```
tvatter: Same question?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -771,7 +874,12 @@ Vinecop::simulate(const size_t n,
                   const std::vector<int>& seeds) const
 {
   auto u = tools_stats::simulate_uniform(n, d_, qrng, seeds);
-  return inverse_rosenblatt(u, num_threads);
+  // inverse_rosenblatt() only works for continous models
+  auto actual_types = var_types_;
+  set_continuous_var_types();
```
tvatter: Can you explain why setting to continuous before sampling is the right thing to do?

tvatter: (I mean, I get why the inverse rosenblatt only exists for continuous models... but are we actually sampling what we thing that we are sampling when using this method for discrete variables?)

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -1074,14 +1196,30 @@ inline void
...
+    std::stringstream msg;
+    msg << "data has wrong number of columns; "
+        << "expected: " << d_exp << " or " << 2 * d_ << ", actual: " << d_data
+        << " (model contains ";
+    if (d_exp == d_) {
+      msg << "no ";
+    } else {
+      msg << get_n_discrete() << " ";
+    }
+    msg << "discrete variables)." << std::endl;
```
tvatter: Same comment for the plural ;)

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -1157,6 +1295,45 @@ Vinecop::truncate(size_t trunc_lvl)
...
 }
 
+//! set all variable types to continuous.
+//! the function can be const, because var_types_ is mutable.
+inline void
+Vinecop::set_continuous_var_types() const
+{
+  var_types_ = std::vector<std::string>(d_);
+  for (auto& t : var_types_)
+    t = "c";
```
tvatter: Don't you need to adjust the types for all pairs too?

include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -546,25 +567,78 @@ FamilySelector::finalize(size_t trunc_lvl)
 //! @param tree a vine tree.
 //! @return The pseudo-observations for the pair coula, extracted from
 //!     the h-functions calculated in the previous tree.
-inline Eigen::MatrixXd
-VinecopSelector::get_pc_data(size_t v0, size_t v1, const VineTree& tree)
+inline void
+VinecopSelector::add_pc_info(const EdgeIterator& e, VineTree& tree)
```
tvatter: Why is this better?

tvatter: Also, there seem to be a lot of code replication, wouldn't there be a way to modularize a bit?

tvatter: My bad, I saw the old function below :) Still, my comment about modularization stands.

test/src_test/include/test_discrete.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+#include "gtest/gtest.h"
+#include <vinecopulib/bicop/class.hpp>
+#include <vinecopulib/vinecop/class.hpp>
+
+namespace test_discrete {
+
+using namespace vinecopulib;
+
+TEST(discrete, bicop)
```
tvatter: What annoys me is that we don't really check that the pdf/hfuncs/hinvs are correct, but I reckon that it's kinda hard to do...

include/vinecopulib/bicop/implementation/kernel.ipp
tnagler: accident

include/vinecopulib/bicop/implementation/kernel.ipp
tnagler: should be no difference, but it's getting replaced by `tools_eigen::trim` anyway

include/vinecopulib/bicop/implementation/kernel.ipp
tnagler: Oh good catch, I started the branch before https://github.com/vinecopulib/vinecopulib/commit/89696890395fc1083faaf01a6119c93a98106cec.

include/vinecopulib/bicop/implementation/kernel.ipp
tnagler: Let's talk about the purely discrete case for simplicity. The copula "pdf" is 
> p(u, v) := C(u, v) - C(u^-, v) - C(u, v^-) + C(u^-, v^-)

Now, the (true) jittered density is constant in the square [u^-,u] x [v^-, v] with value p(u, v). If we estimate it, we will have the least bias in the center of this square. That's what I compute in the specialized pdf. Computing the sum/difference of the cdfs would be computationally more demanding and less accurate.

The reasoning is the same for hfuncs and discrete-continuous pairs.

include/vinecopulib/bicop/implementation/tll.ipp
tnagler: Yeah, we should document it. We should probably even make it completely deterministic, either using a uniform grid or some sort of fixed ghalton grid (asymptotics will be the same as random jittering in any case).

include/vinecopulib/bicop/implementation/tll.ipp
tnagler: I'll change it for consistency

include/vinecopulib/bicop/implementation/tll.ipp
tnagler: NaNs are removed [here](https://github.com/vinecopulib/vinecopulib/blob/discrete-no-dc/include/vinecopulib/bicop/implementation/class.ipp#L699).

include/vinecopulib/bicop/implementation/tll.ipp
tnagler: See above

include/vinecopulib/vinecop/class.hpp
tnagler: Because we need to create a copy inside the function anyway.

include/vinecopulib/misc/triangular_array.hpp
tnagler: Accident

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: That's what the clang-format does with Mozilla style.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: I'll change all of them.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Hm not sure, the vine object hasn't seen any data...

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: - Yeah, it should by in the try block. 
- A `try` always needs to be followed by a `catch`. Otherwise it doesn't compile.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: See above

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Agree

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Agree

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: I'll change to `natural_types` and `natural_array`

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Sorry, don't understand your question.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Yeah, we could always do a check if both variables are continuous, but that makes the code a lot uglier. I would assume that copying the data has has almost impact on the performance, but we can measure.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Yes, the copula is always a continuous distribution. We just need to evaluate it differently to compute likelihoods and such. The sample should always be continuous; discreteness comes in when we transform with discrete margins.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: yep

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Indeed

test/src_test/include/test_discrete.hpp
tnagler: Yeah, we can remove these parts if you prefer. The test just made it a lot easier to identify bugs during development.

Most stuff is checked implicitly though:
- `Bicop::pdf()` when fitting `Bicop` models,
- `Bicop::hfunc1/2()`  when fitting `Vinecop` models,
- `Bicop::hinv1/2()` always uses numerical inversion when discrete variables are present.

include/vinecopulib/bicop/implementation/class.ipp
tvatter: Don't care actually, it's just because we used `size_t` everywhere. Just one thing: maybe you need a static cast or CRAN will complain...

include/vinecopulib/vinecop/implementation/tools_select.ipp
tnagler: Done

include/vinecopulib/vinecop/class.hpp
tvatter: Then you can remove `&`, but why remove the `const` ?

include/vinecopulib/vinecop/implementation/class.ipp
tvatter: ok :)

include/vinecopulib/vinecop/class.hpp
tnagler: Because instead of creating a copy explicitly, we just modify the copy created implictly^^

include/vinecopulib/misc/implementation/tools_eigen.ipp
tvatter: This is what I was thinking about:

```
inline Eigen::VectorXd
unique(const Eigen::VectorXd& x)
{
  // Creates an empty hashset
  std::unordered_set<double> s;
  
  size_t n = x.size();
  double tmp;
  // Traverse the input vector
  for (size_t i = 0; i < n; i++) {
    tmp = x(i);
    // If not present, then put it in the hashtable
    if (s.find(tmp) == s.end()) {
      s.insert(tmp);
    }
  }
  std::vector<double> v(s.begin(), s.end());
  return Eigen::Map<Eigen::VectorXd>(&v[0], v.size());
}
```

include/vinecopulib/vinecop/implementation/class.ipp
tvatter: For `hfunc2_sub.col(j)`, you use the column `d_ + disc_cols[order[j] - 1]` of `u` if the variable is discrete, and the column `order[j] - 1` if the variable is continuous. I get why you actually need `d_ + disc_cols[order[j] - 1]` for the discrete, but why `order[j] - 1` for continuous?

include/vinecopulib/vinecop/implementation/class.ipp
tvatter: Right, thanks!

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: Ah yeah, that's not necessary (since I initialize to 0 above). I'll remove it.

tnagler:
Merging for now, the redundant copies were added to #438.

PR #457, opened 2019-10-17
Partial selection again

tvatter:


include/vinecopulib/vinecop/class.hpp
```diff
@@ -81,12 +81,16 @@ class Vinecop
...
+              const FitControlsVinecop& controls = FitControlsVinecop());
+
+  void DEPRECATED
+  select_all(Eigen::MatrixXd data,
+             const FitControlsVinecop& controls = FitControlsVinecop());
 
-  void select_families(
-    Eigen::MatrixXd data,
-    const FitControlsVinecop& controls = FitControlsVinecop());
+  void DEPRECATED
```
tnagler: we use `DEPRECATED void` in `FitControlsVinecop` any reason for doing it the other way around?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -305,13 +306,13 @@ Vinecop::make_pair_copula_store(const size_t d, const size_t trunc_lvl)
 //!   observations, where \f$ k \f$ is the number of discrete variables.
 //! @param controls the controls to the algorithm (see FitControlsVinecop).
 inline void
-Vinecop::select_all(Eigen::MatrixXd data,
-                    const FitControlsVinecop& controls)
+Vinecop::select(Eigen::MatrixXd data, const FitControlsVinecop& controls)
```
tnagler: The new behavior needs proper documentation.

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -320,6 +321,32 @@ Vinecop::select_all(Eigen::MatrixXd data,
...
+//!
+//! @param data \f$ n \times (d + k) \f$ or \f$ n \times 2d \f$ matrix of
+//!   observations, where \f$ k \f$ is the number of discrete variables.
+//! @param controls the controls to the algorithm (see FitControlsVinecop).
+inline void
+Vinecop::select_all(Eigen::MatrixXd data, const FitControlsVinecop& controls)
+{
+  select(data, controls);
+}
+
```
tnagler: Shouldn't we set `vine_struct_ = RVineStructure(tools_stl::seq_int(1, d_), static_cast<size_t>(0))` first?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -340,20 +367,7 @@ inline void
...
-    if (controls.needs_sparse_select()) {
-      selector.sparse_select_all_trees(data);
-      vine_struct_ = selector.get_rvine_structure(); // can be truncated
-    } else {
-      selector.select_all_trees(data);
-    }
-    finalize_fit(selector);
-  }
+  select(data, controls);
 }
```
tnagler: and here maybe `controls.set_trunc_lvl(vine_struct_.get_trunc_lvl())` to maintain the old behavior.

include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -97,9 +97,20 @@ inline VinecopSelector::VinecopSelector(const Eigen::MatrixXd& data,
...
+}
+
+inline VinecopSelector::VinecopSelector(const Eigen::MatrixXd& data,
+                                        const RVineStructure& vine_struct,
+                                        const FitControlsVinecop& controls,
+                                        std::vector<std::string> var_types)
+  : VinecopSelector(data, controls, var_types)
+{
+  vine_struct_ = vine_struct;
+  proximity_ = false;
```
tnagler: I wonder whether there's a better name than `proximity`. Maybe something like `structure_provided_` or `tree_fixed`?

include/vinecopulib/vinecop/tools_select.hpp
```diff
@@ -205,12 +213,10 @@ class StructureSelector : public VinecopSelector
...
-  void add_allowed_edges(VineTree& tree);
-
   void finalize(size_t trunc_lvl);
 };
 
-class FamilySelector : public VinecopSelector
+class FamilySelector : public StructureSelector
 {
 public:
   FamilySelector(const Eigen::MatrixXd& data,
```
tnagler: I guess that all can go now.

test/src_test/include/test_vinecop_class.hpp
```diff
@@ -406,9 +411,9 @@ TEST_F(VinecopTest, sparse_both_selection)
...
   // controls.set_show_trace(true);
   Vinecop fit(7);
-  fit.select_all(u, controls);
+  fit.select(u, controls);
   EXPECT_NEAR(fit.get_loglik(), fit.loglik(u), 0.001);
-  fit.select_families(u, controls);
+  fit.select(u, controls);
   EXPECT_NEAR(fit.get_loglik(), fit.loglik(u), 0.001);
 }
 }
```
tnagler: Should we have a dedicated unit test for partial selection (checking that the specified trees are actually in the final model)?

include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -517,34 +667,6 @@ StructureSelector::finalize(size_t trunc_lvl)
...
-        pc_data.leftCols(2), tree_criterion, controls_.get_weights());
-      vine_tree[e].weight = w;
-      vine_tree[e].crit = crit;
-    }
-  }
-}
-
 inline void
 FamilySelector::finalize(size_t trunc_lvl)
 {
```
tnagler: Should be removed.

include/vinecopulib/vinecop/implementation/tools_select.ipp
tvatter: Meh, you really like those better? I don't really care...

test/src_test/include/test_vinecop_class.hpp
tvatter: My thought was that I really replicated the tests that we had before, plus I added the test that the partial selection behaves as expected in `fixed_truncation`. Not sure what we would gain from the dedicated unit test, but I'm open if you can make a case!

include/vinecopulib/vinecop/implementation/tools_select.ipp
tnagler: It's not that I like the others so much, but `proximity_` is at best obscure:
- The proximity condition needs to be satisfied no matter if `proximity_` is true or false.
- When `proximity_ == false`, the proximity condition is always true and we don't have to check for it.
- The proximity condition is really not the important aspect in this whole thing.

So I think we should find a name that better conveys the meaning of this flag. When reading the `if` switches, I think we mean
```
if (the structure in this tree is unknown) {
  // loop through all possible edges
} else { 
  // use the known edges
}
```
so the flag's name should optimally convey this. Other suggestions: `structure_known`, `tree_structure_known`, `select_edges`, ... I don't really care about the specific name as long as it can be understood without understanding the code inside the if switches.

test/src_test/include/test_vinecop_class.hpp
tnagler: If I understand correctly, the unit tests currently only show that we can do what we did before without bugs. But I don't think that the new feature is tested, i.e. if I specify only the first 4 trees, does the fitted model actually contain those first 4 trees + more?

include/vinecopulib/vinecop/implementation/tools_select.ipp
tvatter: I like `structure_known`, I push with this name..

test/src_test/include/test_vinecop_class.hpp
tvatter: This is tested here: https://github.com/vinecopulib/vinecopulib/pull/457/files#diff-3d4a12377af54307871de500e34d1849R372-R373

tvatter: https://github.com/vinecopulib/vinecopulib/pull/457/commits/80eaf29cde9a7cbc525c51b421aeaab1b38e26b1

PR #465, opened 2019-11-08
Better constructors

tvatter:
- `Bicop`:
    - [x] add `var_types`
- `Vinecop`
    - [x] add `var_types` 
    - [x] remove checks for matrix inputs
    - [x] remove constructors with order and array
    - [x] collapse constructors using structure only with those using structure + pcs

### DIFF
--- a/include/vinecopulib/bicop/implementation/class.ipp
+++ b/include/vinecopulib/bicop/implementation/class.ipp
@@ -577,9 +577,9 @@ Bicop::set_var_types(const std::vector<std::string>& var_types)
   var_types_ = var_types;
   if (bicop_) {
     bicop_->set_var_types(var_types);
-  }
-  if (tools_stl::is_member(static_cast<size_t>(rotation_), { 90, 270 })) {
-    flip_var_types();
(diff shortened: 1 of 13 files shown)


include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -560,20 +569,15 @@ Bicop::set_parameters(const Eigen::MatrixXd& parameters)
...
-    if (!tools_stl::is_member(t, { "c", "d" })) {
-      throw std::runtime_error("var type must be either 'c' or 'd'.");
-    }
-  }
+  check_var_types(var_types);
   var_types_ = var_types;
-  bicop_->set_var_types(var_types);
+  if (bicop_) {
+    bicop_->set_var_types(var_types);
+  }
```
tnagler: I think the `flip_var_types()` below also needs to be in this block as it modifies the `bicop_` object-

tnagler:
I fixed a few minor issues, otherwise I'm very happy with the PR. You can merge if you agree with my modifications.

PR #471, opened 2019-11-13
rename version.hpp to avoid clashes with other libs

tnagler:


cmake/findHeaders.cmake
```diff
@@ -92,4 +92,4 @@ file(GLOB_RECURSE bicop_hpp ${vinecopulib_includes}/vinecopulib/bicop/*.hpp)
 file(GLOB_RECURSE vinecop_hpp ${vinecopulib_includes}/vinecopulib/vinecop/*.hpp)
 file(GLOB_RECURSE misc_hpp ${vinecopulib_includes}/vinecopulib/misc/*.hpp)
 file(GLOB_RECURSE main_hpp ${vinecopulib_includes}/vinecopulib.hpp)
-file(GLOB_RECURSE version_hpp ${vinecopulib_includes}/version.hpp)
+file(GLOB_RECURSE version_hpp ${vinecopulib_includes}/vinecopulib-version.hpp)
```
tvatter: Is it standard?

tnagler: I barely couldn't find any project that has a header like this. But what we currently have is certainly problematic. If you had another project with a similar structure in the include path, `#include "version.hpp"` would give a random result. Alternatively, we could move `version.hpp` and `mainpage.h` into `include/vinecopulib/` (maybe that's better)

tvatter: OK, so a brief recap: I followed the boost convention for that stuff, and if you notice, for both boost and vinecopulib, the file is installed in install_folder/boost/version.hpp or install_folder/vinecopulib/version.hpp, where install_folder is e.g. /usr/include or /usr/local/include. As such, there is no potential conflict, since you would always need to include the file via #include "vinecopulib/version.hpp"... Or am I missing something?

tnagler: Ok so why don't we just put the file in "vinecopulib/version.hpp" right away. If you just include the headers without a install, there will be a problem.

tvatter: While I agree with what your suggestion, I think that there was a good reason for the current structure, but I honestly don't remember ^^ Maybe not.

PR #473, opened 2019-11-14
fix natural order in RVineStructure::simulate()

tnagler:
We changed what 'natural order' means, but the algorithm is for
the old version.


tvatter:
Good catch!

PR #474, opened 2019-11-14
Fix warnings and make strict compilation part of CI

tvatter:


include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -592,21 +594,19 @@ VinecopSelector::add_pc_info(const EdgeIterator& e, VineTree& tree)
...
     return vertex_data.hfunc1;
   } else {
     return vertex_data.hfunc2;
   }
 }
 
 inline Eigen::VectorXd
-VinecopSelector::get_hfunc_sub(const VertexProperties& vertex_data,
-                               unsigned short pos)
+VinecopSelector::get_hfunc_sub(const VertexProperties& vertex_data, bool pos)
```
tnagler: Maybe we could name the argument `pos0` or `is_first` or something to make it more clear. I was a bit thrown off because it looked like you reversed the logic.

PR #499, opened 2020-05-28
Allow 1-dimensional models

tnagler:
Have done some additional tests in R already: https://github.com/vinecopulib/rvinecopulib/tree/1dim


tvatter:
Sorry for the naive question, but what's the point?

tnagler:
- Came up in a context, where I want to stitch together several vines, possibly 1-dimensional. 
- More generally, I can imagine situation where the dimension of the data is not known upfront and could by 1. In that case, user's would need to if-switch around this case everywhere in the code.
- Don't really see a reason to exclude d = 1 since it's super easy to handle internally.

(codacy activated)

PR #503, opened 2020-07-03
Bicop copy ctor/assignment

tnagler:
Not sure if this was intentional behavior (but I don't think so):
``` cpp
auto rho = Eigen::VectorXd::Constant(1, 0.5);
Bicop bc1(BicopFamily::gaussian, 0, rho);
std::cout << "bc1 is: " << bc1.str() << std::endl;
Bicop bc2 = bc1;
bc2.set_parameters(rho.array() + 0.2);
std::cout << "bc1 is: " << bc1.str() << std::endl;
```
Output (before PR):
```
bc1 is: Gaussian, parameters = 0.5
bc1 is: Gaussian, parameters = 0.7
```
Output (after PR):
```
bc1 is: Gaussian, parameters = 0.5
bc1 is: Gaussian, parameters = 0.5
```

### DIFF
--- a/include/vinecopulib/vinecop/implementation/class.ipp
+++ b/include/vinecopulib/vinecop/implementation/class.ipp
@@ -364,27 +364,13 @@ Vinecop::select_families(const Eigen::MatrixXd& data,
 inline Bicop
 Vinecop::get_pair_copula(const size_t tree, const size_t edge) const
 {
-  if (tree > d_ - 2) {
-    std::stringstream message;
-    message << "tree index out of bounds" << std::endl
-            << "allowed: 0, ..., " << d_ - 2 << std::endl
-            << "actual: " << tree << std::endl;
-    throw std::runtime_error(message.str().c_str());
-  }
-  if (edge > d_ - tree - 2) {
-    std::stringstream message;
-    message << "edge index out of bounds" << std::endl
-            << "allowed: 0, ..., " << d_ - tree - 2 << std::endl
-            << "actual: " << edge << std::endl
-            << "tree level: " << tree << std::endl;
-    throw std::runtime_error(message.str().c_str());
-  }
+  this->check_indices(tree, edge);
   if (tree >= pair_copulas_.size()) {
-    // vine is truncated
-    return Bicop();
+    return Bicop(); // vine is truncated
   }
   return pair_copulas_[tree][edge];
 }
+
 
 //! @brief Gets all pair copulas.
 //!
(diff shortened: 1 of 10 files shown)


include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -780,8 +782,8 @@ Vinecop::pdf(Eigen::MatrixXd u, const size_t num_threads) const
         tools_interface::check_user_interrupt(edge % 100 == 0);
         // extract evaluation point from hfunction matrices (have been
         // computed in previous tree level)
-        Bicop edge_copula = get_pair_copula(tree, edge);
-        auto var_types = edge_copula.get_var_types();
+        Bicop* edge_copula = &pair_copulas_[tree][edge];
```
tvatter, line 785: I like the idea of not copying, but maybe a smart pointer rather than a raw pointer?

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -1070,7 +1072,7 @@ Vinecop::rosenblatt(const Eigen::MatrixXd& u, const size_t num_threads) const
         }
 
         // h-functions are only evaluated if needed in next step
-        Bicop edge_copula = get_pair_copula(tree, edge).as_continuous();
+        Bicop edge_copula = pair_copulas_[tree][edge].as_continuous();
```
tvatter, line 1075: Shouldn't it also be a pointer here?

tvatter:
Oh wow, good catch!

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: good idea

tnagler: Actually it is not. The 'benefit' of a smart pointer would be that the pointed-to object is destroyed when the pointer goes out of scope, but that's not what we want here. The pointed-to object is destroyed whenever `pair_copulas_` is destroyed.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: As it is now, `as_continuous()` always makes exactly one copy. If variables are already continuous, the pair copula is just returned as a copy. If it is discrete, a temporary copy is created inside the function and then automatically moved into `edge_copula` (see [Return Value Optimization](https://en.wikipedia.org/wiki/Copy_elision#:~:text=In%20the%20context%20of%20the,a%20function's%20return%20value). So a pointer wouldn't save us any memory here.

tvatter:
Random failures?

tnagler:
Yep

PR #513, opened 2020-08-14
Fix Frank Kendall's tau

tnagler:
### Issue
I was made aware of a bug in the Frank tau/par conversion. Notice the discontinuity at `x = 3` (which corresponds to the split of cases in our code):
``` r
sq <- seq(2.95, 3.05, by = 0.02)
rbind(sq, sapply(sq, function(x) par_to_ktau("frank", 0, x)))
#>        [,1]      [,2]      [,3]      [,4]      [,5]      [,6]
#> sq 2.9500000 2.9700000 2.9900000 3.0100000 3.0300000 3.0500000
#>      0.1121638 0.1129257 0.1136878 0.3081188 0.3098588 0.3115938
```

### Solution
I fixed the bug in the e67791b. Compared to the [implementation](
https://github.com/openturns/openturns/blob/073d803c3bc8e6ec5cea9d38b79d00610098aab3/lib/src/Base/Func/SpecFunc/Debye.cxx) we probably took this from, we declare a `const kk` which should in fact be modified in the loop (in the `x < 3` case).  After the fix:
``` r
sq <- seq(2.95, 3.05, by = 0.02)
rbind(sq, sapply(sq, function(x) par_to_ktau("frank", 0, x)))
#>        [,1]      [,2]      [,3]      [,4]      [,5]      [,6]
#> sq 2.9500000 2.9700000 2.9900000 3.0100000 3.0300000 3.0500000
#>    0.3028686 0.3046237 0.3063738 0.3081188 0.3098588 0.3115938
```

### More
Since the code is so ugly, I took a few minutes to simplify and clean up things. The code now only computes the Debye order 1 function (which is what we need) instead of general nth order. I also moved it into the vinecopulib namespace.

### Additional checks
``` r
sq <- seq(-10, 10, by = 0.01)
taus_copula <- sapply(sq, function(x) copula::tau(copula::frankCopula(x)))
taus_vcl <- sapply(sq, function(x) par_to_ktau("frank", 0, x))
all.equal(taus_copula, taus_vcl)
#>  [1] TRUE
```


include/vinecopulib/bicop/implementation/frank.ipp
```diff
@@ -102,198 +97,125 @@ FrankBicop::get_start_parameters(const double tau)
...
- Engeln et al, Colloid & Polymer Sci. 261 (9) (1983) 736
- Maximon , Proc. R. Soc. A 459 (2039) (2003) 2807
- @param[in] x the argument and upper limit of the integral. x>=0.
- @param[in] n the power in the numerator of the integral, 1<=n<=20 .
- @return the Debye function. Zero if x<=0, and -1 if n is outside the
-     parameter range that is implemented.
- @author Richard J. Mathar
- @since 2007-10-31 implemented range n=8..10
-*/
+
```
tvatter, line 101: The doc appears twice

tvatter:
Wild bug, good catch! The doc of the debye function appears twice, otherwise good to go, but I must admit that I didn't check the implementation details (i.e., formulas).

PR #514, opened 2020-09-24
Discrete fix

tnagler:
Output of pdf/loglik wasn't correct when discrete variables are present. This was caused by unitialized columns that were used later in the algorithm. Also clarified what we call "copula density" in that case, because there's no standard definition.


tvatter:
Sorry, but why `2 + (u.cols() == 4) * disc_col` ?

tnagler:
We already know that there is one discrete and one continuous variable. Now there are two cases:

1. `u.cols() == 3`: then the F(x^-) values for the discrete variable is always in the last column, i.e. `u.col(2)`.

2. `u.cols() == 4`: Then the F(x^-) values for the discrete variable is in the third column if variable 1 is discrete, and in the fourth column if variable 2 is discrete. Thus, `u.col(2 + disc_col)`.

Let me just copy&paste this into the code as a comment to avoid confusion in the future.

PR #519, opened 2020-10-14
fix user-level var types in flip()

tnagler:
Fixes https://github.com/vinecopulib/rvinecopulib/issues/226


tvatter:
LOL, good catch!

PR #523, opened 2020-11-12
fix threshold selection issue with independent data

tnagler:
fixes https://github.com/vinecopulib/rvinecopulib/issues/228

### DIFF
--- a/include/vinecopulib/vinecop/implementation/fit_controls.ipp
+++ b/include/vinecopulib/vinecop/implementation/fit_controls.ipp
@@ -155,11 +155,7 @@ FitControlsVinecop::get_trunc_lvl() const
 inline void
 FitControlsVinecop::set_trunc_lvl(size_t trunc_lvl)
 {
-  if (trunc_lvl == 0) { // select level (likely passed from R)
-    trunc_lvl_ = std::numeric_limits<size_t>::max();
-  } else {
-    trunc_lvl_ = trunc_lvl;
-  }
+  trunc_lvl_ = trunc_lvl;
 }
 
 //! returns whether to select the truncation level automatically.
--- a/include/vinecopulib/vinecop/implementation/tools_select.ipp
+++ b/include/vinecopulib/vinecop/implementation/tools_select.ipp
@@ -182,23 +182,7 @@ VinecopSelector::sparse_select_all_trees(const Eigen::MatrixXd& data)
   double d = static_cast<double>(d_);
 
   std::vector<double> thresholded_crits;
-  if (controls_.get_select_threshold()) {
-    // initialize threshold with maximum pairwise |tau| (all pairs get
-    // thresholded)
-    auto tree_crit = controls_.get_tree_criterion();
-    auto crits =
-      calculate_criterion_matrix(data, tree_crit, controls_.get_weights());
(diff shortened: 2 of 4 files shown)


include/vinecopulib/vinecop/implementation/fit_controls.ipp
```diff
@@ -155,11 +155,7 @@ FitControlsVinecop::get_trunc_lvl() const
 inline void
 FitControlsVinecop::set_trunc_lvl(size_t trunc_lvl)
 {
-  if (trunc_lvl == 0) { // select level (likely passed from R)
```
tvatter, line 158: Why do you remove that?

include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -182,23 +182,7 @@ VinecopSelector::sparse_select_all_trees(const Eigen::MatrixXd& data)
   double d = static_cast<double>(d_);
 
   std::vector<double> thresholded_crits;
-  if (controls_.get_select_threshold()) {
```
tvatter, line 185: Same question: why do you remove that?

include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -349,7 +335,7 @@ inline double
 VinecopSelector::get_next_threshold(std::vector<double>& thresholded_crits)
 {
   if (thresholded_crits.size() == 0) {
-    return 0.0;
```
tvatter, line 352: LOL, I really don't understand anything at this PR, sorry. Can you explain?

include/vinecopulib/vinecop/implementation/fit_controls.ipp
tnagler: Oh right that was a bit unrelated, but appears to be a relict from old times. I doesn't seem we're ever passing 0 from R when we mean Inf (we probably did in an earlier version though). However, nowadays we have proper support for 0-truncated vines if I remember correctly (which we didn't in the beginning). I changed this during debugging and then forgot about it.

include/vinecopulib/vinecop/implementation/tools_select.ipp
tnagler: I initially thought it was smart to warm start the threshold search with a value calculated form the unconditional Kendall's tau. But the problem is that the very first model that we try might already be worse than an independence model. In that case, the algorithm stopped without having any model at hand, which caused a segfault. Now we simply start with threshold = 1 (which fits an independence model) and then start a proper search.

include/vinecopulib/vinecop/implementation/tools_select.ipp
tnagler: - threshold = 1.0 for starting the search as explained above.
- `double mbicv_opt = std::numeric_limits<double>::max()` is because the independence model has mbicv > 0, so the check for `mbicv >= mbicv_opt` at the end of the iteration would also terminate the search after the first iteration. This fixes it.

tnagler: And yeah, this code isn't very pretty in general.

PR #530, opened 2021-01-06
Fix tll npars

tnagler:
`npars` was default initialized to 0 after copying a `KernelBicop` object.


tvatter:
Somewhat unrelated, but get_npars should be a const method.

PR #563, opened 2022-08-30
replace bitwise ops on boolean operands

tnagler:


tvatter:
What's that about?

tnagler:
https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/rvinecopulib-00check.html

tnagler:
But good that you ask, since this flag only exists for clang-14 and up ;)

PR #565, opened 2022-12-04
Replace sprintf

tnagler:
- this is necessary because `sprintf()` is now flagged as potential security  risk by CRAN checks, which issue a WARNING.
 - in our case, `boost/odeint` makes (safe) use of the function, but we don't really need it.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -43,7 +43,7 @@ jobs:
           fi
         shell: bash
       - name: Setup-R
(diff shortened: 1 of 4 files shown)


tnagler:
I'll do that locally in `tools_integration.hpp` now, where the problem occurs. Since macOS deprecated `sprintf` in version 13, this should also be relevant outside of CRAN checks. 

In case boost [fixes this](https://github.com/boostorg/odeint/pull/58) (last commit to [boost/odeint ](https://github.com/boostorg/odeint) is 4 years ago), we can remove it.

PR #586, opened 2024-10-23
CMake: Fix interface includes

jschueller:


tvatter:
Hi, I've switched the base branch to `dev`, because this is what we use for PRs before making a full release to `main` (which is admittedly long overdue). We've recently revamped our CI workflows so I'm not surprised this causes issues.

tvatter:
@jschueller I've fixed the dependencies install CI issue. Now, it looks like your changes do not compile on windows. Can you explain a bit more what you are trying to fix? [Dev branch](https://github.com/vinecopulib/vinecopulib/tree/dev) seemed to compile fine just last week. And we've released a new version of pyvinecopulib (i.e., the Python API) using the dev branch from vinecopulib 2 days ago without issue.

jschueller:
indeed the dev branch doesnt have this c++14 compilation issue, but now it just have an issue with eigen & wdm includes not being propagated through cmake config

cmake/buildTargets.cmake
```diff
@@ -9,10 +9,12 @@ if (VINECOPULIB_SHARED_LIB)
 else()
     add_library(vinecopulib INTERFACE)
     target_include_directories(vinecopulib INTERFACE
-            $<BUILD_INTERFACE:${vinecopulib_includes}>
-            $<INSTALL_INTERFACE:include/vinecopulib>)
```
tvatter, line 13: Why is this removal needed?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.13)
```
tvatter, line 1: I remember having to upgrade the minimum required cmake at some point. Version 3.2 is from 2015, is this requirement an issue?

tvatter:
@jschueller Thanks a lot for the PR. I just left two really quick comments. My knowledge of cmake is limited to this project, apologies if those aren't relevant.

cmake/buildTargets.cmake
```diff
@@ -9,10 +9,12 @@ if (VINECOPULIB_SHARED_LIB)
...
+            $<BUILD_INTERFACE:${vinecopulib_includes}>)
 endif()
 
+target_include_directories (vinecopulib INTERFACE $<INSTALL_INTERFACE:include>)
+target_link_libraries(vinecopulib INTERFACE Eigen3::Eigen wdm)
```
tvatter, line 16: Why is the link to boost not needed then?

jschueller: indeed lets add boost too

cmake/buildTargets.cmake
jschueller: its not removed: I kept the BUILD_INTERFACE part here,
 then moved the INSTALL_INTERFACE outside of the if because its also needed for the library
note that I had to remove the /vinecopulib subdirectory because vinecopulib.hpp is at the root dir

jschueller:
finally!

examples/bicop/CMakeLists.txt
```diff
@@ -9,24 +9,21 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/bin)
...
-
-# Set required variables for includes and libraries
-# In the second line
-#   * VINECOPULIB_LIBRARIES is needed if vinecopulib has been built as a
-#     shared lib (does nothing otherwise).
-#   * CMAKE_THREAD_LIBS_INIT is needed for some linux systems
-#     (but does nothing on OSX/Windows).
-set(external_includes ${VINECOPULIB_INCLUDE_DIR} ${EIGEN3_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${wdm_INCLUDE_DIRS})
-set(external_libs ${VINECOPULIB_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
+find_package(vinecopulib REQUIRED)
```
tvatter, line 26: This seems like dark magic. So we don't need to explicitly find and include the dependencies anymore when using vinecopulib ?

examples/bicop/src/CMakeLists.txt
```diff
@@ -1,9 +1,5 @@
-# Include header files
-include_directories(${external_includes})
```
tvatter, line 2: Sorry, asking for my personal understanding: why don't we need the external includes anymore? Are they automatically included by the "dark magic" above?

examples/bicop/CMakeLists.txt
jschueller: yes!

cmake/options.cmake
```diff
@@ -2,9 +2,10 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS "ON")
...
 
 option(CMAKE_BUILD_TYPE "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." "Release")
-option(VINECOPULIB_SHARED_LIB    "Pre-compiled version"              "OFF")
-option(WARNINGS_AS_ERRORS        "Compiler warnings as errors"       "OFF")
-option(OPT_ASAN                  "Use adress sanitizer (debug)"       "ON")
-option(BUILD_TESTING             "Build tests."                       "ON")
-option(CODE_COVERAGE             "Code coverage."                    "OFF")
-option(STRICT_COMPILER           "Stricter compiler warnings"        "OFF")
\ No newline at end of file
+option(VINECOPULIB_SHARED_LIB    "Pre-compiled version"              OFF)
```
tvatter, line 5: Ah, so CMake actually has boleans, I thought one had to use strings instead. Good to know.

examples/bicop/src/CMakeLists.txt
jschueller: yes, they are automatically added when you link to vinecopulib

tvatter:
@jschueller Can you explain briefly what this PR does and/or what problem it solves? Both @tnagler and myself are super grateful for anything that improves the build system (thanks!!!) as we've always been struggling with it. But since what is currently in dev "works", and I'm always hesitant to fix what ain't broken, I think that some context would be helpful.

jschueller:
Its just that: it simplifies the use of vinecopulib from another cmake project to avoid the need to specify third-party dependencies.

tvatter:
Got it, awesome, thanks again, merging!

jschueller:
are there plans for a new release of vinecopulib that would include these changes ?

tvatter:
@tnagler ?

tvatter:
@jschueller After checking with @tnagler, there are a few additional features he wants to include in the next release, and he mentioned that he would be able to make a PR for those in the next few weeks. Is there a specific reason you need a quick release to main? Apologies if we haven't released officially for a while and I agree that we should do so ASAP, thanks again for your PR :)

jschueller:
its just that the current version does not even build, but it can wait of course

tnagler:
If urgent, we could also release as is and patch the additional feature before we do the update of the R interface. @tvatter ?

tvatter:
@tnagler  If the zi-stuff can be ready in the next week or two, then release can wait. Otherwise I think that it would be a good idea to just go ahead given that (a) what's in main is severely outdated, (b) pyvinecopulib already uses the latest from dev, (c) we can't know for sure when the zi-stuff will be ready.

tvatter:
Hi @jschueller, sincere apologies. Both @tnagler and myself have a nightmarish kind of semester and have been swamped by other things. We need to update & regenerate the doc because there hasn't been a release in a while. Realistically, I will personally only be able to work on this next month, after I'm done with teaching. Maybe @tnagler has some availability before ? But I wouldn't get my hopes up. Really sorry, it's not that I don't want to.

cmake/options.cmake
```diff
@@ -2,9 +2,10 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS "ON")
...
+option(BUILD_SHARED_LIBS         "shared/static lib"                 ON)
```
tvatter, line 6: @jschueller I just noticed this one. Why is this needed? It doesn't seem to be used anywhere in the code base.

jschueller: its used by add_library if the latter does not specify SHARED or STATIC explicitely

tvatter: But we already have a "VINECOPULIB_SHARED_LIB",  should those two be separated ?

tvatter: (sorry again, we're cmake noobs)

jschueller: VINECOPULIB_SHARED_LIB enables the precompiled library, but BUILD_SHARED_LIBS decides if its shared or static

tvatter: Got it, thanks a lot!

jschueller: but maybe VINECOPULIB_SHARED_LIB should be renamed

tvatter: good idea, I opened an issue

PR #597, opened 2024-12-10
improve discrete tll from rvinecopulib@pull-zi

tnagler:


include/vinecopulib/misc/implementation/tools_stats.ipp
```diff
@@ -61,6 +62,38 @@ simulate_uniform(const size_t& n,
...
+//!
+//! @param n Number of observations.
+//! @param d Dimension.
+//! @param seeds Seeds of the random number generator; if empty (default),
+//!   the random number generator is seeded randomly.
+//!//!
+//! @return An \f$ n \times d \f$ matrix of independent
+//! \f$ \mathrm{N}(0, 1) \f$ random variables.
+inline Eigen::MatrixXd
+simulate_normal(const size_t& n, const size_t& d, std::vector<int> seeds)
```
tvatter, line 75: Shouldn't this implementation rely on `simulate_uniform` and `qnorm` ? Then it's literally 2 lines of code, avoid plenty of copy-pasting, offers the quasi-rng possiblity, etc

include/vinecopulib/misc/tools_stats.hpp
```diff
@@ -52,6 +54,52 @@ qnorm(const Eigen::MatrixXd& x)
...
 }
 
+//! @brief Quantile function of the Standard normal distribution with additional
+//! bound checks for numerical stability.
+//!
+//! @param x Evaluation points.
+//!
+//! @return An \f$ n \times d \f$ matrix of evaluated quantiles.
+inline Eigen::MatrixXd
+safe_qnorm(const Eigen::MatrixXd& x)
```
tvatter, line 64: * The important part is only =0 or =1 right? Because why would one evaluate quantiles at probabilities outside [0,1]?
* Any reason you don't rely on Eigen's ndtri?

include/vinecopulib/misc/tools_stats.hpp
```diff
@@ -52,6 +54,52 @@ qnorm(const Eigen::MatrixXd& x)
...
+}
+
+//! @brief Distribution function of the Standard normal distribution with
+//! additional bound checks for numerical stability.
+//!
+//! @param x Evaluation points.
+//!
+//! @return An \f$ n \times d \f$ matrix of evaluated quantiles.
+inline Eigen::MatrixXd
+safe_pnorm(const Eigen::MatrixXd& x)
```
tvatter, line 87: * Same question as above regarding the bounds.
* Shouldn't you use the formula along with std::erf ?

include/vinecopulib/misc/implementation/tools_stats.ipp
```diff
@@ -145,6 +178,48 @@ to_pseudo_obs_1d(Eigen::VectorXd x, const std::string& ties_method)
   return x / (static_cast<double>(n) + 1.0);
 }
 
+//
```
tvatter, line 181: Needs documentation

include/vinecopulib/misc/tools_stats.hpp
```diff
@@ -100,12 +148,109 @@ simulate_uniform(const size_t& n,
...
+                const size_t& d,
+                std::vector<int> seeds = std::vector<int>());
+
 Eigen::VectorXd
 to_pseudo_obs_1d(Eigen::VectorXd x, const std::string& ties_method = "average");
 
 Eigen::MatrixXd
 to_pseudo_obs(Eigen::MatrixXd x, const std::string& ties_method = "average");
 
+class BoxCovering
```
tvatter, line 162: * Most of this should go in the ipp file.
* Needs documentation

include/vinecopulib/misc/tools_stats.hpp
```diff
@@ -100,12 +148,109 @@ simulate_uniform(const size_t& n,
...
+{
+public:
+  explicit BoxCovering(const Eigen::MatrixXd& u, uint16_t K = 40)
```
tvatter, line 165: Why uint16_t ? Shouldn't we use the same kind as in the grid for kernel bicops ?

tvatter:
Most of my comments are cosmetic, because as you said functionally we know it works, and we'll rethink performance in due time. One thing that is missing however: unit tests.

tnagler:
It still runs throught the TLL/discrete tests we hade before:
https://github.com/vinecopulib/vinecopulib/blob/258eb2d7758ce067ee7d53789067f0d3ebbb9375/test/src_test/include/test_discrete.hpp#L70

I know that's not great, but also not sure what exactly to test here.

include/vinecopulib/misc/tools_stats.hpp
tnagler: This saves many `static_cast<>`s when multiplying `K` or `K_` with `double` - in lines where we already cast the result of floor/ceil back to `size_t`. Simply multiplying `size_t * double` gave a STRICT_COMPILER fail on some of our CI builds. I can do the casts if you prefer.

include/vinecopulib/misc/tools_stats.hpp
tnagler: it's actually even better. our new qnorm/pnorm variants already do proper handling of the edge cases, so the safe variants can be removed.

include/vinecopulib/misc/implementation/tools_stats.ipp
tnagler: done

include/vinecopulib/misc/tools_stats.hpp
tnagler: done

include/vinecopulib/misc/tools_stats.hpp
tnagler: see below

include/vinecopulib/misc/implementation/tools_stats.ipp
tvatter: Is this the computational bottleneck?

tvatter:
Maybe at least some quick sanity checks? copilot is pretty good at inventing those directly in vscode... so is chatgpt but with some copy-pasting.

include/vinecopulib/misc/implementation/tools_stats.ipp
tnagler: Nah, overhead is small anyway.

tnagler:
Ok added some tests.

PR #598, opened 2024-12-11
Remove boost from dpqnorm

tvatter:
On my computer, this gives a 25% speedup for the tll benchmark. Would appreciate if @tnagler can confirm on another machine. Implementation is cleaner anyway, so unless there's a regression, I think it's a good thing.


tnagler:
On my laptop specifically it's about 5% slower now. But I agree that it's cleaner and would expect speedups on most machines.

PR #600, opened 2024-12-12
Show the Eigen version that's found

tvatter:


jschueller:
@tnagler you could just use the cmake variable Eigen3_VERSION, no need to parse the header

PR #603, opened 2024-12-14
Cross-platform random numbers and jitter variant for pseudo-obs

tvatter:
* [make random numbers cross-platform reproducible](https://github.com/vinecopulib/vinecopulib/commit/866a05897d4b1ad57337caa18647de9b871511ce)
* [add fixed jitter variant for pseudo_obs](https://github.com/vinecopulib/vinecopulib/commit/9188a5d2a950d32266c144541ae53120309323d8)


tvatter:
@tnagler  apparently windows isn't happy (LINK : fatal error LNK1104: cannot open file 'libboost_random-vc143-mt-gd-x64-1_84.lib' )

include/vinecopulib/misc/implementation/tools_stats.ipp
```diff
@@ -140,6 +142,49 @@ to_pseudo_obs_1d(Eigen::VectorXd x,
   return x.array() / (static_cast<double>(n) + 1.0);
 }
 
+//! @param ties_method Indicates how to treat ties; same as in R, see
+//! https://stat.ethz.ch/R-manual/R-devel/library/base/html/rank.html.
+//! @return Pseudo-observations of the copula, i.e. \f$ F_X(x) \f$
+//! (column-wise).
+inline Eigen::MatrixXd
```
tvatter, line 149: After thinking about it, shouldn't this belong to wdm since you moved all the related code there?

tvatter:
To complete this, I think we should pull wdm 0.2.5 and modify the cmake build to include the USE_BOOST flag in the same place the other boost flags are included. Is that correct @tnagler ?

include/vinecopulib/misc/implementation/tools_stats.ipp
tnagler: right this function is redundant now, removed it.

PR #606, opened 2024-12-17
Switch to quickpool

tvatter:
After doing from the root

```
mkdir build && cd build && cmake .. -DBUILD_TESTING=OFF && make && sudo make install
```

And then, after doing

```
mkdir build && cd build && cmake .. && make && cd .. && rm -rf build && ./bin/main && rm -rf bin/
```

From `examples/benchmark` with `examples/benchmark/src/main.cpp` (i.e., sequential result):

image

Then, adding the following line 59:

```
config.second.set_num_threads(8); // Set threads to 8 for benchmarking
```

Before:

image

After:

image

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -70,7 +70,7 @@ jobs:
           fi
         shell: bash
       - name: Code coverage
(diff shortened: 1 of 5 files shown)


tnagler:
We'll revisit this later.

PR #620, opened 2025-01-11
Fix tawn

tnagler:


include/vinecopulib/bicop/family.hpp
```diff
@@ -127,12 +127,6 @@ const std::vector<BicopFamily> itau = {
   BicopFamily::joe
 };
 
-//! Families that can be flipped by adjusting the rotation.
-const std::vector<BicopFamily> flip_by_rotation = {
```
tvatter, line 131: Why remove this?

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -731,22 +732,17 @@ Bicop::get_var_types() const
 inline void
 Bicop::flip()
```
tvatter, line 733: Sorry, I don't understand what's going on anymore. Why is the rotation changed irrespective of whether a family can be flipped by rotation? If we call `bicop_->flip();` anyway, should we really change the rotation for families which shouldn't be flipped by rotation? And it was the case before already, but why no flip by rotation for the case 0/180?

tnagler: The "flip by rotation family" logic was misleading. The families not included don't allow for rotation != 0, so they aren't affected by changing the rotation anyway. 

For families that are exchangeable (most in our lib), flipping 0 or 180 deg rotated versions does not alter the shape of the copula. For all others, it would not be a meaningful adjustment because it is unrelated to exchanging arguments.

We aleays change 90 <-> 270 because the definition of these rotations involves exchanging arguments plus a reflection in one of the arguments. When "flipping" (=changing order of arguments), which argument is reflected needs to change also. For families that are exchangeable in their 0deg variant, this will be the only change this is needed.

bicop_->flip() then implements any changes to the shape beyond this. For most of our families, it does nothing.

tnagler: added a comment along these lines to the code

include/vinecopulib/bicop/family.hpp
tnagler: see below

PR #626, opened 2025-01-15
CMake: Fix dependent target detection

jschueller:
find_dependency must occur before the include else configure fails


tnagler:
changed base to dev, we will have a release to main today or tomorrow.

tvatter:
Thanks @jschueller

Issue #627, opened 2025-01-15
rotation selection


jschueller:
would it be possible to control the allowed rotation selection in FitControls settings ?
one could want to discard 180°-rotated copulas to prefer 0° copulas
for example my data it gives a 180° Joe instead of a 0° Clayton
While I could set the family to clayton only, I could want to allow only 0° copulas from all families
or allow all but just prefer 0° copulas over other rotated candidates when aic/bic values are close

tvatter:
@tnagler requested something similar here: https://github.com/vinecopulib/rvinecopulib/issues/279

Seems reasonable to me. One thing that I wonder is about the 90/270 given that the archimedean we have only allow for positive dependence when non-rotated.

@tnagler do you have a preference on how to proceed?

tnagler:
Yeah I'm not quite sure how to do this right. How about we add a `allowed_rotations` control with default `{0, 90, 180, 270}`? 

Stuff like "if aic are close, then prefer ..." I find too vague (and niche) to implement in generality.

tvatter:
I'm fine with this, and it includes your allow_rotation=False with allowed_rotations={0}

tvatter:
See #628 and let me know what you think

tvatter:
True, yes. I went for this, and it's unit tested.

tnagler:
Fixed by #628, already included in the last release.

PR #628, opened 2025-01-16
Rotations option in fit controls

tvatter:
See #627


tnagler:
looks good, can you add an entry in NEWS.md?

PR #629, opened 2025-01-16
Fit controls config

tvatter:


test/src_test/include/test_bicop_sanity_checks.hpp
```diff
@@ -91,6 +91,33 @@ TEST(bicop_sanity_checks, controls_checks)
...
+  EXPECT_EQ(controls.get_nonparametric_method(), controls2.get_nonparametric_method());
+  EXPECT_EQ(controls.get_nonparametric_mult(), controls2.get_nonparametric_mult());
+  EXPECT_EQ(controls.get_selection_criterion(), controls2.get_selection_criterion());
+  EXPECT_EQ(controls.get_weights(), controls2.get_weights());
+  EXPECT_EQ(controls.get_psi0(), controls2.get_psi0());
+  EXPECT_EQ(controls.get_preselect_families(), controls2.get_preselect_families());
+  EXPECT_EQ(controls.get_allow_rotations(), controls2.get_allow_rotations());
+  EXPECT_EQ(controls.get_num_threads(), controls2.get_num_threads());
+}
+
```
tnagler, line 120: Should be better to set and test against non-default values of the parameters, otherwise we might miss something just because the defaults are equal in both constructors.

test/src_test/include/test_vinecop_sanity_checks.hpp
```diff
@@ -58,6 +58,33 @@ TEST(vinecop_sanity_checks, controls_print)
...
+  EXPECT_EQ(controls.get_tree_criterion(), controls2.get_tree_criterion());
+  EXPECT_EQ(controls.get_threshold(), controls2.get_threshold());
+  EXPECT_EQ(controls.get_select_threshold(), controls2.get_select_threshold());
+  EXPECT_EQ(controls.get_select_trunc_lvl(), controls2.get_select_trunc_lvl());
+  EXPECT_EQ(controls.get_select_families(), controls2.get_select_families());
+  EXPECT_EQ(controls.get_show_trace(), controls2.get_show_trace());
+  EXPECT_EQ(controls.get_num_threads(), controls2.get_num_threads());
+  EXPECT_EQ(controls.get_mst_algorithm(), controls2.get_mst_algorithm());
+}
+
```
tnagler, line 87: See previous comment.

test/src_test/include/test_bicop_sanity_checks.hpp
tvatter: Is that alright now?

test/src_test/include/test_vinecop_sanity_checks.hpp
tvatter: Is that alright now?

PR #633, opened 2025-02-10
CMake: Find Boost in CONFIG mode

jschueller:
avoids the cmake warning:
```
CMake Warning (dev) at /usr/share/cmake/Modules/CMakeFindDependencyMacro.cmake:76 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning
```


tvatter:
Thanks @jschueller

PR #635, opened 2025-03-05
fix TLL speed issues related to FFT (fixes #304)

tnagler:


tvatter:
Modifying `examples/benchmark/main.cpp` lines 131-132:

```
  //benchmark_vinecop_fitting();
  benchmark_bicop_tll(500);
  benchmark_bicop_tll(499);
```

image

Solves the issue indeed (was 5ms vs 50ms before)!

PR #644, opened 2025-05-14
Add RVineTrees

tvatter:
Allow to

* merge multiple vines
* fill incomplete vines

Note that

* Codacy is tripping again
* Coverage misses 
    * in rvine_trees.ipp are related to defensive error throwing in private members, no idea how to trigger those, they shouldn't happen,
    * in tools_stl.hpp are IMO codecov tripping because the related lines are obviously covered.


include/vinecopulib/vinecop/rvine_trees.hpp
```diff
@@ -0,0 +1,68 @@
...
+    : d_(0)
+    , trunc_lvl_(0)
+    , trees_() {};
+
+  explicit RVineTrees(const std::vector<size_t>& order,
+                      const TriangularArray<size_t>& struct_array);
+  explicit RVineTrees(const std::vector<RVineTrees>& vines);
+
+  const std::vector<Tree>& get_trees() const { return trees_; }
+
```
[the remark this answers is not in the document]
tvatter: In vinecopulib, doxygen comments are placed in .ipp files. Can you update?

include/vinecopulib/vinecop/implementation/rvine_trees.ipp
```diff
@@ -0,0 +1,419 @@
...
+//
+// This file is part of the vinecopulib library and licensed under the terms of
+// the MIT license. For a copy, see the LICENSE file in the root directory of
+// vinecopulib or https://vinecopulib.github.io/vinecopulib/.
+
+namespace vinecopulib {
+
+inline RVineTrees::RVineTrees(
+  const std::vector<size_t>& order,
+  const TriangularArray<size_t>& struct_array) // TODO: natural order
```
tnagler, line 11: TODO still relevant?

include/vinecopulib/vinecop/implementation/rvine_trees.ipp
```diff
@@ -0,0 +1,419 @@
...
+        C.insert(struct_array(k, e));
+      }
+      tree.push_back(Edge(a, b, C));
+    }
+    trees_[t] = tree;
+  }
+}
+
+inline std::tuple<std::vector<size_t>, TriangularArray<size_t>>
+RVineTrees::to_struct_array(bool fill_missing) const
```
tnagler, line 38: it's a bit weird that `to_struct_array()` returns a order/struct_array-tuple. how about making it `to_rvine_structure()` and returning `RVineStructure()` (which is effectively order+struct_array)?

tnagler (commented):
Looks good overall, but currently it's not used for anything. It's also sufficiently separated from the main code to be part of an external library as far as I can see. That would change if we replace `boost/graph` with this of course...

include/vinecopulib/vinecop/implementation/rvine_trees.ipp
tvatter: The reason is that there's the fill_missing option. If `false` and the underlying RVineTrees isn't complete, then one can't convert to an `RVineStructure`.

include/vinecopulib/vinecop/implementation/rvine_trees.ipp
tvatter: No!

tvatter:
> Looks good overall, but currently it's not used for anything. It's also sufficiently separated from the main code to be part of an external library as far as I can see. That would change if we replace `boost/graph` with this of course...

Well, it's not "currently", but I'm working on this vines merging stuff, and this is the basic infrastructure. I think that it's better to break PRs into manageable chunks. I'm also preparing something related to sampling orders, and one implementation uses this.

PR #656, opened 2025-08-25
Refactor controls

tvatter:


include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -880,19 +966,20 @@ Bicop::fit(const Eigen::MatrixXd& data, const FitControlsBicop& controls)
 //!
 //! @param data An \f$ n \times (2 + k) \f$ matrix of observations contained in
 //!   \f$(0, 1) \f$, where \f$ k \f$ is the number of discrete variables.
-//! @param controls The controls (see `FitControlsBicop`).
+//! @param controls The controls (see `FitControls`).
 inline void
-Bicop::select(const Eigen::MatrixXd& data, FitControlsBicop controls)
+Bicop::select(const Eigen::MatrixXd& data, FitControls& controls)
```
tnagler, line 971: Wouldn't it be better to just pass the controls by value/copy? The current version is modifying the user's controls while the function is running. Seems too risky for a function that is likely called in parallel.

include/vinecopulib/misc/fit_controls.hpp
```diff
@@ -7,81 +7,163 @@
...
-//! Configuration options for initializing a FitControlsVinecop object.
+namespace vinecopulib {
+
+//! @brief controls options for bivariate and vine copula models.
 //!
-//! This struct provides a flexible way to initialize a `FitControlsVinecop`
-//! object. Each field is optional, and default values are applied if the field
-//! is not set.
-struct FitControlsConfig
+//! @details This struct provides a flexible way to controlsure copula fitting
```
tnagler, line 27: find&replace error?

include/vinecopulib/vinecop/class.hpp
```diff
@@ -40,18 +40,44 @@ class Vinecop
...
     const Eigen::MatrixXd& data,
-    const Eigen::Matrix<size_t, Eigen::Dynamic, Eigen::Dynamic>& matrix =
-      Eigen::Matrix<size_t, Eigen::Dynamic, Eigen::Dynamic>(),
-    const std::vector<std::string>& var_types = {},
-    const FitControlsVinecop& controls = FitControlsVinecop());
+    const Eigen::Matrix<size_t, Eigen::Dynamic, Eigen::Dynamic>& matrix,
+    const std::vector<std::string>& var_types = {});
+  explicit Vinecop(
+    const Eigen::MatrixXd& data,
+    FitControls& controls,
```
tnagler, line 61: Also here: passing controls by reference is dangerous.

tnagler (commented):
Two more fundamental concerns:
* Passing the controls by reference avoids copies, but is really dangerous when the functions are called in parallel. I think we should pass by const reference where possible, and by value otherwise.
* It's weird that we're mostly using the struct with all the optionals effectively just as if there are no optionals and everything is available. Superficially, this looks like adding a large number of `.value()` calls with no clear purpose. If the `std::optional`s are only there to allow for named parameters in the constructor, maybe there's a way to only use them in a special constructor, but keep the `struct` with regular values?

tvatter:
> * Passing the controls by reference avoids copies, but is really dangerous when the functions are called in parallel. I think we should pass by const reference where possible, and by value otherwise.

Yes, I'm aware. My thinking was, we're fully in control of what happens with the object in the parallel parts of the library. The only place I've identified to be a potential issue is [here](https://github.com/vinecopulib/vinecopulib/pull/656/files#diff-e0aa12fc3eeefd47a496dfafad0dd1dc52fd97e234d54e093a24bd4bac0093d5R1066), and I specifically create a copy.

But I agree that this is not completely satisfying. It probably speaks in favor of including this `validated` flag to the struct. Basically, we use const ref everywhere, and just (a) copy and validate/set whenever `validated=false`, or (b)  trust the provided controls whenever `validated=true`. 

> * It's weird that we're mostly using the struct with all the optionals effectively just as if there are no optionals and everything is available. Superficially, this looks like adding a large number of `.value()` calls with no clear purpose. If the `std::optional`s are only there to allow for named parameters in the constructor, maybe there's a way to only use them in a special constructor, but keep the `struct` with regular values?

There are two goals:

* Avoid the named parameters as mentioned.
* Avoid carrying around unnecessary parameters.

For now, the second goal is only used when working with Bicop (all the additional Vinecop controls are neither set nor used). Another place I think we want this behavior is [here](https://github.com/vinecopulib/vinecopulib/pull/656/files#diff-6c9d21a7d681eb433681e788eb804e283ec1fbd5495b58a82c284c259de3530bR885-R889). It's not part of this PR, but I would like to only pass this struct there, and one wouldn't need e.g. the parametric_method if the family is nonparametric (and vice versa). Using optionals allow for this flexibility everywhere.

PR #664, opened 2026-04-10
fix typo in pdf_d_d, adjust threshold for analytical derivatives

tnagler:


include/vinecopulib/vinecop/implementation/tools_select.ipp
```diff
@@ -483,7 +484,7 @@ VinecopSelector::select_edges(VineTree& vine_tree)
     boost::mt19937 gen = controls_.get_rng();
 
     // Randomize root vertex
-    std::uniform_int_distribution<size_t> root_dist(0, d - 1);
+    boost::random::uniform_int_distribution<size_t> root_dist(0, d - 1);
```
tvatter, line 487: Any reason?

tnagler: didn't compile on my machine. some versions of boost and stdlib used by the compiler are incompatible (mismatch between what is declared/expected as constexpr)

tvatter: OK!

PR #674, opened 2026-06-23
Allow a custom tree_criterion function for vine structure selection

tvatter:
Add a "custom" tree_criterion to FitControlsVinecop that selects the vine structure with a user-supplied edge-weight function (std::function<double(const MatrixXd&, const VectorXd&)>) instead of only the built-in dependence measures. Reachable via
FitControlsVinecop::set_tree_criterion_function and FitControlsConfig::tree_criterion_function, and threaded through tools_select::calculate_criterion.

Extracted from #657 without the #656 controls refactor it was based on. Also give the FitControlsVinecop members in-class defaults so that construction from a partial FitControlsConfig no longer leaves them uninitialized.

Will allow us to close [this pyvinecopulib issue too](https://github.com/vinecopulib/pyvinecopulib/issues/203)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,12 @@
 
 ### NEW FEATURES
 
+* Allow a user-supplied edge-weight function for vine structure selection via
+  `tree_criterion = "custom"` together with
+  `FitControlsVinecop::set_tree_criterion_function` (or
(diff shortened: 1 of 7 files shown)


tnagler (approved):
LGTM

PR #675, opened 2026-06-24
Evaluate bivariate copulas with per-row parameters

tvatter:
## Per-row parameters for `Bicop` (+ eval-core refactor)

Adds overloads of `Bicop::pdf`, `cdf`, `hfunc1`, `hfunc2`, `hinv1`, `hinv2`, and `loglik` that evaluate a parametric bivariate copula at a **different parameter set per row of `u`** in one (optionally multi-threaded) call. Following review, it also refactors the evaluation core to a single parameter-aware interface and fixes a latent out-of-bounds bug in the mixed discrete/continuous h-inverse.

```cpp
Eigen::VectorXd pdf(const Eigen::MatrixXd& u,
                    const Eigen::MatrixXd& parameters,  // n x p, one set per row
                    size_t num_threads = 1) const;
```

### What's in it
- **Feature — per-row parameters.** Additive overloads for all six evaluators + `loglik`, parametric families only (TLL throws; independence accepts an `n×0` matrix). The facade validates shape/bounds/finiteness and parallelizes over row batches.
- **Refactor — one leaf per primitive.** Each of `pdf_raw/cdf/hfunc1_raw/hfunc2_raw/hinv1_raw/hinv2_raw` is now a single **parameter-aware pure-virtual** leaf; the duplicated no-argument state forms are removed. State dispatchers call the leaves with the stored parameters as a broadcast row; nonparametric families ignore the argument and use their grid. Net **−235 lines**.
- **Bug fix — mixed-discrete h-inverse.** For a `{d,c}` copula with a numeric inverse (Frank, BB*, Tawn), `hinv2` stripped to two columns and the bisection re-entered the discrete difference-quotient branch of `hfunc1` on a 2-column matrix → out-of-bounds read (heap corruption in release, assertion in debug). The numeric inverse is split so the continuous `_raw` primitives invert the continuous `*_raw` leaves; the discrete branch is now reachable only with 4-column data by construction, and the `u.cols()==4` guard is removed.
- **Internal convention.** The internal parameter layout now matches the public API (`n×p`, one row per observation), so `format_parameters` no longer transposes.

### Correctness
The h-inverse is correct for all four var-type configs: stripping the discrete companion column when the *conditioning* variable is continuous is **exact** (`{d,c}` hinv2 ≡ `{c,c}` hinv2 bit-for-bit), and the difference-quotient inversion for discrete-conditioning configs is the right primitive. Confirmed from first principles, against VineCopula's `BiCopHinv` (~3e-11), and by bit-identity + round-trip tests.

### Tests & verification
Per-row parity (all methods × families × rotations), discrete configs, threading, broadcast, NaN/bounds, `{d,c}`≡`{c,c}` bit-identity, continuous round-trip. Release **350/350**; Debug **+ASan 350/350** (no sanitizer errors); `STRICT_COMPILER` GNU build clean; `VINECOPULIB_PRECOMPILED=ON` build + tests pass; R-parity; clang-format-14 clean.

### Downstream
No public-API break (leaves are `protected`); `rvinecopulib`/`pyvinecopulib` keep compiling. Exposing the overloads is a pin bump + bindings.

### DIFF
--- a/include/vinecopulib/bicop/abstract.hpp
+++ b/include/vinecopulib/bicop/abstract.hpp
@@ -67,7 +67,7 @@ protected:
   // following are virtual so they can be overriden by KernelBicop
   virtual Eigen::VectorXd pdf(const Eigen::MatrixXd& u);
 
-  virtual Eigen::VectorXd cdf(const Eigen::MatrixXd& u) = 0;
+  virtual Eigen::VectorXd cdf(const Eigen::MatrixXd& u);
 
   virtual Eigen::VectorXd hfunc1(const Eigen::MatrixXd& u);
 
@@ -77,23 +77,12 @@ protected:
 
   Eigen::VectorXd hinv2(const Eigen::MatrixXd& u);
 
-  virtual Eigen::VectorXd pdf_raw(const Eigen::MatrixXd& u) = 0;
-
-  virtual Eigen::VectorXd hfunc1_raw(const Eigen::MatrixXd& u) = 0;
-
-  virtual Eigen::VectorXd hfunc2_raw(const Eigen::MatrixXd& u) = 0;
-
-  virtual Eigen::VectorXd hinv1_raw(const Eigen::MatrixXd& u) = 0;
-
-  virtual Eigen::VectorXd hinv2_raw(const Eigen::MatrixXd& u) = 0;
-
-  // Parameter-aware overloads. `parameters` has shape p x m with m in {1, n}:
-  // column j holds the p parameters for observation j; a single column is
-  // broadcast to all rows. These let a copula be evaluated at a different
-  // parameter set per row of `u` without mutating object state. The leaves
-  // (pdf_raw, cdf, hfunc*_raw, hinv*_raw) default to ignoring `parameters` and
-  // using object state, so nonparametric families need no overrides;
-  // parametric families override them with the actual (stateless) math.
+  // Evaluation leaves. `parameters` has shape m x p with m in {1, n}: row i
+  // holds the p parameters for observation i; a single row is broadcast to all
+  // observations. The state-based dispatchers above call these with the stored
+  // parameters (a 1 x p broadcast row). This is the sole evaluation interface;
+  // every family implements the (stateless) math, and nonparametric families
+  // ignore `parameters` and read their interpolation grid instead.
   Eigen::VectorXd pdf(const Eigen::MatrixXd& u,
                       const Eigen::MatrixXd& parameters);
 
@@ -110,22 +99,22 @@ protected:
                         const Eigen::MatrixXd& parameters);
 
   virtual Eigen::VectorXd cdf(const Eigen::MatrixXd& u,
-                              const Eigen::MatrixXd& parameters);
+                              const Eigen::MatrixXd& parameters) = 0;
 
   virtual Eigen::VectorXd pdf_raw(const Eigen::MatrixXd& u,
-                                  const Eigen::MatrixXd& parameters);
+                                  const Eigen::MatrixXd& parameters) = 0;
 
   virtual Eigen::VectorXd hfunc1_raw(const Eigen::MatrixXd& u,
-                                     const Eigen::MatrixXd& parameters);
+                                     const Eigen::MatrixXd& parameters) = 0;
 
   virtual Eigen::VectorXd hfunc2_raw(const Eigen::MatrixXd& u,
-                                     const Eigen::MatrixXd& parameters);
+                                     const Eigen::MatrixXd& parameters) = 0;
 
   virtual Eigen::VectorXd hinv1_raw(const Eigen::MatrixXd& u,
-                                    const Eigen::MatrixXd& parameters);
+                                    const Eigen::MatrixXd& parameters) = 0;
 
   virtual Eigen::VectorXd hinv2_raw(const Eigen::MatrixXd& u,
-                                    const Eigen::MatrixXd& parameters);
+                                    const Eigen::MatrixXd& parameters) = 0;
 
   virtual Eigen::MatrixXd tau_to_parameters(const double& tau) = 0;
   Eigen::MatrixXd no_tau_to_parameters(const double&);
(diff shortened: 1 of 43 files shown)


include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -394,6 +397,296 @@ Bicop::hinv2(const Eigen::MatrixXd& u) const
...
+  if (!tools_stl::is_member(get_family(), bicop_families::parametric)) {
+    throw std::runtime_error(
+      "per-row parameters are only supported for parametric families; "
+      "nonparametric families store an interpolation grid rather than a "
+      "per-observation parameter vector.");
+  }
+  const Eigen::Index n = u.rows();
+  const Eigen::Index p = get_parameters().rows();
+  if (parameters.rows() != n) {
+    throw std::runtime_error("parameters must have one row per row of u "
```
tnagler, line 633: Isn't the expected format one parameter *column* per row of u?

include/vinecopulib/bicop/implementation/abstract.ipp
```diff
@@ -138,21 +138,151 @@ AbstractBicop::set_var_types(const std::vector<std::string>& var_types)
...
+  }
+  Eigen::VectorXd pdf = pdf_raw(u.leftCols(2));
+  tools_eigen::trim(pdf, DBL_MIN, DBL_MAX);
+  return pdf;
+}
+
+inline Eigen::VectorXd
+AbstractBicop::hfunc1(const Eigen::MatrixXd& u)
+{
+  if (var_types_[0] == "d" && u.cols() == 4) {
```
tnagler: The branch where `var_types_[0] == "d"` but `u.cols() != 4` should never be reachted or there must be a bug somewhere else.

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -394,6 +397,296 @@ Bicop::hinv2(const Eigen::MatrixXd& u) const
...
+                                      const Eigen::MatrixXd&)>& f) const
+{
+  const size_t n = static_cast<size_t>(u.rows());
+  Eigen::VectorXd out(n);
+  if (n == 0) {
+    return out;
+  }
+  auto do_batch = [&](const tools_batch::Batch& b) {
+    out.segment(b.begin, b.size) = f(u.middleRows(b.begin, b.size),
+                                     parameters_t.middleCols(b.begin, b.size));
```
tnagler: Here we take one column per row for example..

include/vinecopulib/bicop/implementation/clayton.ipp
```diff
@@ -19,66 +19,78 @@ inline ClaytonBicop::ClaytonBicop()
...
   return (std::pow(u, -theta) - 1) / theta;
 }
 
 inline double
-ClaytonBicop::generator_inv(const double& u)
+ClaytonBicop::generator_inv(const double& u,
+                            const Eigen::Ref<const Eigen::VectorXd>& parameters)
 {
-  double theta = double(this->parameters_(0));
+  double theta = double(parameters(0));
```
tnagler: While this is not all new to this PR, I'm wondering what the `double()` is supposed to achieve here and many other places (some new). `parameters(0)` should already be `double`.

include/vinecopulib/bicop/implementation/gaussian.ipp
tnagler: We now have the same code more or less twice in this family. Shouldn't the `xxx_raw(u)` variant just call `xxx_raw(u, parameters_)` as for the other families?

tnagler: More generally, why do we even keep the separate `xxx_raw(u)` variant around?

include/vinecopulib/bicop/implementation/student.ipp
```diff
@@ -95,6 +86,93 @@ StudentBicop::hinv1_raw(const Eigen::MatrixXd& u)
   return hinv;
 }
 
+inline Eigen::VectorXd
+StudentBicop::pdf_raw(const Eigen::MatrixXd& u)
```
tnagler: Redundant code here too.

include/vinecopulib/bicop/abstract.hpp
```diff
@@ -87,6 +87,46 @@ class AbstractBicop
...
+  Eigen::VectorXd hfunc2(const Eigen::MatrixXd& u,
+                         const Eigen::MatrixXd& parameters);
+
+  Eigen::VectorXd hinv1(const Eigen::MatrixXd& u,
+                        const Eigen::MatrixXd& parameters);
+
+  Eigen::VectorXd hinv2(const Eigen::MatrixXd& u,
+                        const Eigen::MatrixXd& parameters);
+
+  virtual Eigen::VectorXd cdf(const Eigen::MatrixXd& u,
```
tnagler, line 99: I think we should remove the variants without `parameters` entirely. The virtuals with `parameters` can take default argument `Eigen::MatrixXd()`.

tnagler (changes requested):

include/vinecopulib/bicop/abstract.hpp
tvatter: Done — the no-arg `*_raw(u)` virtuals are gone; each primitive is a single parameter-aware pure virtual implemented once per family. I kept them *pure* rather than using a default `Eigen::MatrixXd()` argument: a default arg on a virtual binds to the static type, which is a footgun since everything is called through `AbstractBicop*`. The state dispatchers (`pdf(u)`, …) pass the stored parameters as a broadcast row instead.

include/vinecopulib/bicop/implementation/student.ipp
tvatter: Removed entirely (per the above) — Gaussian and Student now have only the parameter-aware leaves. Gaussian's `pdf_raw` keeps a `parameters.rows()==1` scalar (Cholesky) branch so the common single-parameter/state path stays **bit-identical** and allocation-free.

include/vinecopulib/bicop/implementation/gaussian.ipp
tvatter: Removed entirely (per the above) — Gaussian and Student now have only the parameter-aware leaves. Gaussian's `pdf_raw` keeps a `parameters.rows()==1` scalar (Cholesky) branch so the common single-parameter/state path stays **bit-identical** and allocation-free.

include/vinecopulib/bicop/implementation/clayton.ipp
tvatter: Removed — all redundant `double(...)` casts on parameter indexing are gone (36 across the families). Kept the one genuine `1×1`→scalar cast in `tll.ipp`.

include/vinecopulib/bicop/implementation/abstract.ipp
tvatter: It *was* reachable, via a real bug. For a `{d,c}` copula with a numeric inverse, `hinv2` strips to two columns (continuous conditioning) and the bisection re-entered the discrete branch of `hfunc1` on a 2-column matrix → out-of-bounds. Fixed at the root: the continuous `_raw` numeric inverse now inverts the continuous `*_raw` leaf, so the discrete branch is only reached with 4-column data. The `&& u.cols()==4` guard is removed.

include/vinecopulib/bicop/implementation/class.ipp
tvatter: The public API stays `n×p` (one parameter set per *row*, matching the downstream request). What you spotted was the internal `p×n` and the transpose in `format_parameters` — I've flipped the internal convention to `n×p` too, so public and internal now match and the transpose is gone (broadcast is detected by `parameters.rows()==1`).

PR #682, opened 2026-07-10
Add tail dependence and Blomqvist's beta to Bicop

tvatter:
## Summary

Adds two dependence-measure methods to `Bicop`, mirroring VineCopula's `BiCopPar2TailDep` and `BiCopPar2Beta`:

- **`parameters_to_taildep(parameters)` / `get_taildep()`** — tail dependence coefficients, returned as a **2×2 `Eigen::MatrixXd`** covering all four corners of the unit square.
- **`parameters_to_beta(parameters)` / `get_beta()`** — Blomqvist's beta.

## The 2×2 tail-dependence matrix

`M(i, j)` is the coefficient as `U1 → i`, `U2 → j`, with `i, j ∈ {0, 1}` (0 = lower, 1 = upper):

|          | `U2 → 0` | `U2 → 1` |
| -------- | -------- | -------- |
| `U1 → 0` | `M(0,0)` = lower (λ_L) | `M(0,1)` = upper-left |
| `U1 → 1` | `M(1,0)` = lower-right | `M(1,1)` = upper (λ_U) |

The diagonal entries are the classical lower/upper coefficients (VineCopula's `$lower`/`$upper`). The full matrix is needed because a 2-value pair cannot represent the off-diagonal corners, which are non-zero for:

- **90°/270°-rotated families** — dependence sits in the upper-left / lower-right corners.
- **the Student t copula** — has tail dependence in *all four* corners (the discordant corners use `ρ → −ρ`).

## Implementation

- **Beta**: one generic implementation in `AbstractBicop` via `β = 4·C(0.5, 0.5) − 1` (works for every family, including the nonparametric `tll`), sign-flipped for 90°/270° rotations in the `Bicop` wrapper — same rule as Kendall's τ.
- **Tail dependence**: closed forms per parametric family; a zero-matrix default in `AbstractBicop` covers `indep`/`gaussian`/`frank`. `tll` returns `NaN` (no closed-form limit). The `Bicop` wrapper rotates the 2×2 matrix *as a grid, counter-clockwise*, to match the data rotation in `rotate_data`.
- Uses dev's stateless `cdf(u, parameters)` / `pickands(t, par)` evaluation interface (post-#675), so no state save/restore is needed.

## Tests

- Extends the R parity test to cross-check `get_beta()` and the diagonal of `get_taildep()` against `BiCopPar2Beta`/`BiCopPar2TailDep` for **all parametric families × 4 rotations**.
- Adds `test_bicop_taildep` for what VineCopula can't check: the off-diagonal corners under rotation, Student-t four-corner symmetry, the β sign-flip, getter/method consistency, and the `tll` `NaN` contract.
- Full suite: **358/358 pass**; clang-format-14 clean.

## Notes

- Additive, non-breaking public API. Downstream `rvinecopulib` / `pyvinecopulib` wrappers will need matching bindings to expose the new methods.

### DIFF
--- a/include/vinecopulib/bicop/implementation/abstract.ipp
+++ b/include/vinecopulib/bicop/implementation/abstract.ipp
@@ -97,12 +97,14 @@ AbstractBicop::no_tau_to_parameters(const double&)
   throw std::runtime_error("Method not implemented for this family");
 }
 
-//! Default tail dependence: no tail dependence in any of the four corners.
-//! Families with a closed form override this.
+//! Default tail dependence: not implemented for this family, so all four
+//! corners are reported as NaN. Families with a closed form override this
+//! (including those that genuinely have zero tail dependence, e.g. `indep`,
+//! `gaussian`, `frank`).
 inline Eigen::MatrixXd
 AbstractBicop::parameters_to_taildep(const Eigen::MatrixXd&)
 {
-  return Eigen::MatrixXd::Zero(2, 2);
+  return Eigen::MatrixXd::Constant(2, 2, NAN);
 }
 
 //! Blomqvist's beta computed generically from the copula cdf as
--- a/include/vinecopulib/bicop/implementation/abstract.ipp
+++ b/include/vinecopulib/bicop/implementation/abstract.ipp
@@ -97,6 +97,29 @@ AbstractBicop::no_tau_to_parameters(const double&)
   throw std::runtime_error("Method not implemented for this family");
 }
 
+//! Default tail dependence: no tail dependence in any of the four corners.
+//! Families with a closed form override this.
+inline Eigen::MatrixXd
+AbstractBicop::parameters_to_taildep(const Eigen::MatrixXd&)
+{
+  return Eigen::MatrixXd::Zero(2, 2);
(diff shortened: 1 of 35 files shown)


include/vinecopulib/bicop/implementation/abstract.ipp
```diff
@@ -97,6 +97,29 @@ AbstractBicop::no_tau_to_parameters(const double&)
   throw std::runtime_error("Method not implemented for this family");
 }
 
+//! Default tail dependence: no tail dependence in any of the four corners.
+//! Families with a closed form override this.
+inline Eigen::MatrixXd
+AbstractBicop::parameters_to_taildep(const Eigen::MatrixXd&)
```
tnagler, line 105: Shouldn't the default be returning NaN for "not implemented" rather than 0 for "no tail dependence"?

tnagler: LGTM otherwise

tvatter: Agreed, modified accordingly.

PR #683, opened 2026-07-10
Analytic derivatives of density/h-functions and analytic Vinecop scores

tvatter:
## What

Analytic derivatives of the bivariate copula density and h-functions, and analytic backends for `Vinecop::scores`, `Vinecop::gradient`, and `Vinecop::hessian`. Feature parity with VineCopula's `BiCopDeriv`/`BiCopDeriv2`/`BiCopHfuncDeriv`/`BiCopHfuncDeriv2` and `RVineGrad`/`RVineHessian`, based on Schepsmeier & Stöber (2014) and Stöber & Schepsmeier (2013).

### `Bicop`

New methods (each with a per-row-parameter overload):
`pdf_deriv`, `pdf_deriv2`, `hfunc1_deriv`, `hfunc1_deriv2`, `hfunc2_deriv`, `hfunc2_deriv2`, `logpdf_deriv`, `logpdf_deriv2`.

- A string selector chooses the derivative direction — `"par1"`, `"par2"`, `"u1"`, `"u2"` (`"par"` is short for `"par1"`); second-order selectors concatenate two, e.g. `"par1u1"`, `"u1u2"` — mirroring VineCopula's `deriv=` argument.
- Rotations are resolved by the chain rule in the facade, so parameters stay in their positive/natural parameterisation (`get_parameters()`).
- Closed forms cover the new `bicop_families::analytic_derivs` group (independence, Gaussian, Student, Clayton, Gumbel, Frank, Joe); every other **parametric** family (BB1/6/7/8, Tawn) falls back to central finite differences inside the leaf, so the whole interface works for all parametric families. Nonparametric families and discrete `var_types` throw.
- Student-t required porting the incomplete-beta derivatives (Boik & Robinson-Cox, 1998) and the t-quantile ν-derivative helpers into `tools_stats`.

### `Vinecop`

- **`scores`** is now analytic: the step-wise scores differentiate each edge's log-density directly, and the full gradient (`step_wise = false`) is an `RVineGrad`-style cascade through the vine.
- **`hessian`** is now analytic in both modes: the joint (`step_wise = false`) Hessian is a second-order cascade (the second derivative of the gradient cascade, i.e. the `RVineHessian` equivalent), and the step-wise Hessian is a first-order cascade of the step-wise score's argument derivatives. Both are built from the `Bicop` leaves, so they cover every parametric family (BB/Tawn via their finite-difference-fallback leaves) and are correct for 90°/270° rotations. Only models with discrete variables keep finite differences.
- **New `scores_full`** exposes the per-edge derivative caches behind the scores (`keep_all` option, mirroring `pdf`/`pdf_full`); `scores` is a thin wrapper over it.
- **New `gradient`** returns the observation-average of the scores (mirroring how `hessian` averages `hessian_full`).
- `scores`/`hessian` now **throw for models with nonparametric (TLL) pair copulas** — differentiating w.r.t. an interpolation grid is meaningless, and the previous code wrote out of bounds (the matrix is sized by effective d.f. but was indexed by the grid values). Discrete variables and every parametric family are unaffected.
- Internally, the gradient, joint Hessian, and step-wise Hessian share one forward pass (`build_deriv_cache`) and one first-order propagation primitive (`propagate_first_order`); `hessian()` aggregates the per-observation Hessian in row-chunks so its peak memory is `O(chunk · npars²)` rather than materialising the full `n · npars²` tensor it only averages (`hessian_full` remains the per-observation entry point).

## Tests

- **Bicop, R-parity**: every derivative method against `BiCopDeriv*` for all families × rotations (`1e-4`).
- **Bicop, self-consistency**: analytic vs. central finite differences of the public (rotation-aware) `pdf`/`hfunc1`/`hfunc2`, all families × rotations — validates the formulas, the rotation chain-rule tables, and the BB/Tawn finite-difference fallback.
- **Vinecop, gradient**: full gradient vs. `RVineGrad` **element-wise** (sum over observations and a single-observation check); threading determinism.
- **Vinecop, Hessian**: analytic joint Hessian vs. brute-force central finite differences of the whole-vine log-likelihood (the ground truth — matches to ~`6e-5` on a mixed 0/90/180/270 + truncated vine), exact symmetry, thread determinism, a BB1-edge vine exercising the fallback leaves, and an **element-wise** parity check against VineCopula's `RVineHessian` directly (no finite differences). The analytic step-wise Hessian is checked against central finite differences of the step-wise scores (including a two-parameter Student edge's own-edge block).

Targeted derivative/scores/Hessian tests pass in Debug (7 vine + 145 bicop parity cases); clang-format-14 clean; `NEWS.md`/`AGENTS.md` updated. (The full suite and the `VINECOPULIB_PRECOMPILED=ON` release build are left to CI.)

## Notes

- Stacked on #679 (base branch `rename-hessian-full`), so it uses the new `hessian`/`hessian_full` names; rebases cleanly onto `dev` once #679 lands.
- **VineCopula cross-checks.** While building the R oracles we found and reported two bugs in VineCopula that had made `RVineGrad`/`RVineHessian` unusable as oracles in some cases: (1) they returned wrong values whenever a 90°/270°-rotated family was in the model (an argument-orientation mismatch between the likelihood forward pass and the derivative cascade); (2) `RVineHessian` ignored the data entirely — the sample size `T` was never assigned and resolved to `TRUE` (= 1). Both are now fixed in **VineCopula 2.6.2** ([tnagler/VineCopula#101](https://github.com/tnagler/VineCopula/issues/101), [#102](https://github.com/tnagler/VineCopula/issues/102)); the tests here require ≥ 2.6.2 and, thanks to the fixes, compare **element-wise and directly** against both `RVineGrad` and `RVineHessian` (the #101 fix is what lets us use `RVineHessian` directly rather than finite-differencing `RVineGrad`).
- **Rotations at the vine level.** The vine-level cross-library parity model deliberately uses only 0°/180° families. vinecopulib and VineCopula do not share the same structure / swapped-rotation conventions, so a vine with 90°/270°-rotated families cannot be constructed identically in both libraries (their log-likelihoods, and hence gradients/Hessians, then differ — a model-construction mismatch, confirmed empirically: the 0/180 model's log-likelihood matches to 6 digits while no rotation-orientation of the rotated model reconciles it). 0/180 families are symmetric under the argument swap that distinguishes the conventions, so both libraries build the *same* model there. Rotated families are covered elsewhere: at the **bicop level** against `BiCopDeriv*` across **all rotations** (the 90°/270° `par1par1` columns previously skipped for the #102 bug are now checked), and at the **vine level** against brute-force finite differences of the whole-vine log-likelihood on a mixed 0/90/180/270 vine.


include/vinecopulib/bicop/implementation/abstract.ipp
```diff
@@ -471,4 +472,219 @@ AbstractBicop::hinv2_num_raw(const Eigen::MatrixXd& u,
...
+  return (c_xy / c - (c_x / c) * (c_y / c)).matrix();
+}
+//! @}
+
+namespace tools_deriv {
+
+//! splits a derivative selector into components; the encoding is the 0-based
+//! parameter index for `"par<k>"`, `-1` for `"u1"`, and `-2` for `"u2"`.
+inline std::vector<int>
+parse_components(const std::string& deriv)
```
tnagler, line 570: Feels like `enum` makes more sense for the derivative selector.

include/vinecopulib/bicop/implementation/gaussian.ipp
```diff
@@ -4,6 +4,8 @@
 // the MIT license. For a copy, see the LICENSE file in the root directory of
 // vinecopulib or https://vinecopulib.github.io/vinecopulib/.
 
+#include <algorithm>
```
tnagler: Don't see what this is used for.

include/vinecopulib/bicop/implementation/parametric.ipp
```diff
@@ -282,5 +282,131 @@ ParBicop::check_fit_method(const std::string& method)
   }
 }
 
+//! @}
+
+//! @name Finite-difference derivative fallbacks
```
tnagler, line 287: Is this the right place? Couldn't this also be useful for some nonparametric methods (esp the u1 and u2 derivs)?

include/vinecopulib/bicop/implementation/student.ipp
```diff
@@ -4,6 +4,8 @@
 // the MIT license. For a copy, see the LICENSE file in the root directory of
 // vinecopulib or https://vinecopulib.github.io/vinecopulib/.
 
+#include <algorithm>
```
tnagler: I get the impression this include should live in tools_eigen?

include/vinecopulib/misc/tools_stats.hpp
```diff
@@ -93,6 +93,69 @@ qt(const Eigen::MatrixXd& x, double nu)
...
+//! \f$ \partial_\nu t_\nu(x) \f$; ported from VineCopula
+//! `tcopuladeriv_new.c` (`diff_dt_nu`).
+double
+diff_dt_nu(double x, double nu);
+
+//! @brief Logarithmic derivative of the Student t density w.r.t. its
+//! argument, \f$ \partial_x \log t_\nu(x) \f$; ported from VineCopula
+//! `tcopuladeriv_new.c` (`diff_dt_u`).
+double
+diff_dt_u(double x, double nu);
```
tnagler: The naming scheme for the t derivatives is a bit inconsistent/I transparent. Would this justify a separate name space? could also be placed in the student implementation files directly.

include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -613,6 +613,540 @@ Bicop::loglik(const Eigen::MatrixXd& u,
...
+//! methods take one component (`"par"` is short for `"par1"`); second-order
+//! methods take two in any order (a single component means differentiating
+//! twice, so `"par"` is short for `"par1par1"`).
+//!
+//! Rotations are handled internally via the chain rule, so derivatives are
+//! always w.r.t. the arguments and (positive, natural) parameters of the
+//! rotated copula. Closed-form expressions are used for the families in
+//! `bicop_families::analytic_derivs`; other parametric families fall back to
+//! central finite differences of the corresponding function. Derivatives
+//! require continuous variable types; nonparametric families throw.
```
tnagler, line 637: Requiring continuous variables is not generally necessary, right? We just haven't implemented the necessary logic for discrete vars..

include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -1276,31 +1758,272 @@ Vinecop::hessian_full(Eigen::MatrixXd u,
...
+            const DerivCache& ce = cache(t, e);
+            size_t msrc = ce.msrc;
+            bool a1 = (aff2[e] != 0);
+            bool a2 = ce.direct ? (aff2[msrc - 1] != 0) : (aff1[msrc - 1] != 0);
+            if (!a1 && !a2) {
+              aff1[e] = 0;
+              aff2[e] = 0;
+              continue;
+            }
+            Eigen::VectorXd v1 = Eigen::VectorXd::Zero(m);
```
tnagler: Don't really get what's happening here.

tnagler: Some more comments what o1, aff2, active etc mean would be helpful

include/vinecopulib/bicop/implementation/gaussian.ipp
tvatter: Removed in e40fbf6 — it was unused here (and in student/gumbel/joe); the real user was `tools_eigen.ipp` (std::min/max/sort/swap), which only got `<algorithm>` transitively. It now includes `<algorithm>`/`<utility>` itself.

include/vinecopulib/bicop/implementation/student.ipp
tvatter: Exactly right — `tools_eigen.ipp` was the actual user and now carries the include itself; the unused copies here and in gaussian/gumbel/joe are removed (e40fbf6).

include/vinecopulib/misc/tools_stats.hpp
tvatter: Renamed in 23638b9 after the `dt`/`pt`/`qt` convention already used in tools_stats, so each name reads as the derivative it computes: `dqt_dnu`, `d2qt_dnu2` (quantile w.r.t. ν), `dpt_dnu`, `d2pt_dnu2` (cdf), `ddt_dx`, `ddt_dnu`, `dlogdt_dx` (density). The doc comments keep the original VineCopula function names as provenance. Kept them in `tools_stats` next to `dt`/`pt`/`qt` rather than a new namespace — happy to move them into the student files instead if you prefer.

include/vinecopulib/vinecop/implementation/class.ipp
tvatter: Added in 74ab24b: the `DerivLeaf`/`DerivCache` structs in class.hpp now explain `o1`/`o2` (the edge's hfunc1/hfunc2 output leaves), `active` (whether a deeper tree consumes that output, from the needed_hfunc masks) and `msrc`/`direct`; the gradient cascade documents `tilde1`/`tilde2` (per-column perturbations of the h-function storage), `aff1`/`aff2` (which columns are affected) and `v1`/`v2` (the edge's argument perturbations), with matching notes for the `hat`/`til`/`bar` state in the Hessian cascades.

tvatter: Done!

include/vinecopulib/bicop/implementation/student.ipp
tvatter: Moved to the right place.

include/vinecopulib/bicop/implementation/gaussian.ipp
tvatter: Moved to the right place.

include/vinecopulib/misc/tools_stats.hpp
tvatter: Agreed, update, let me know what you think

tnagler: Naming is good. I still think it would be better to move this to the student implementation, since this is the only place where all these functions are used.

include/vinecopulib/vinecop/implementation/class.ipp
tnagler: OK now I understand what's going on I think. A few more things for readability:
* I suggest renaming o1/o2 to h1/h2.
* What does `src` in `msrc` stand for? Maybe we can make this more self-explanatory like `arg2_pos` or something.
* This PR re-introduces the "direct/indirect" terminology from VineCopula that we deliberately got rid of in the rest of the library. How about `bool h1_input` or so?
* `DerivCache` has a bit of a weird naming scheme; `DerivLeaf` is clean. Maybe we can just do `dpar_u1` etc? I assume the `l` indicates "log-density", but that's the purpose of the entire `struct`.

include/vinecopulib/bicop/implementation/class.ipp
tvatter: Decision: postpone until the rest is settled and merged.

include/vinecopulib/misc/tools_stats.hpp
tvatter: Done in c3f0e39 — moved `dqt_dnu`/`d2qt_dnu2`/`dpt_dnu`/`d2pt_dnu2`/`ddt_dx`/`ddt_dnu`/`dlogdt_dx`, `inbeder`, and its four continued-fraction coefficient statics into `student.ipp` as inline file-local helpers, and dropped the `tools_stats` declarations (the Boost special-function includes moved with them).

On replacing the continued fraction with Boost/`betaincder`: neither can, because both only give the **x**-derivative of the incomplete beta. Boost's `ibeta_derivative` is ∂I_x/∂x; `betaincder`'s `_betaincderp`/`_betaincderq` are ∂/∂a and ∂/∂b but there is no ∂²/∂a². The t-vs-ν **second** derivatives (`d2pt_dnu2`, and `d2qt_dnu2` through it) need ∂²I_x/∂a², and the plain value I_x is never even used — so the Boik & Robinson-Cox routine stays. The scalar special functions it needs (`lgamma`/`digamma`/`trigamma`/`beta`) already come from Boost. (A `betaincder`-style single-recurrence rewrite extended to ∂²/∂a² would be more compact, but it's new numerical code — a possible future cleanup, not this PR.)

include/vinecopulib/vinecop/implementation/class.ipp
tvatter: Done in 1fd7ee9:
* `o1`/`o2` → `h1`/`h2`.
* `msrc` → `arg2_col` (the storage column that supplies the edge's 2nd argument = its `min_array` entry).
* dropped the `l` prefix: `lu1/lu2/lpar/…` → `du1/du2/dpar/…`. Since `DerivCache` and `DerivLeaf` now share those field names, the struct doc spells out that the cache's are derivatives of `log c` and the leaf's are derivatives of an h-function `h`.
* `direct` → `arg2_is_h2` rather than `h1_input`: `direct == true` selected the source column's **hfunc2**, so `h1_input` would have named the opposite branch and silently flipped the polarity; `arg2_is_h2` keeps the meaning and is self-documenting.

Also cleaned up the cryptic cascade locals (`tilde/aff` → `dh/…_affected`, `hat/til/bar` → `dh_a/dh_b/d2h`, `hp/tp/bp` → `darg1_*`, `Hval` → `H_ab`, etc.) and removed the leftover "direct/indirect" comments. Pure rename, suite still green.

include/vinecopulib/bicop/implementation/parametric.ipp
tvatter: Agreed it belongs at the `AbstractBicop` level. I'll do it in a small follow-up PR on top of this branch rather than here, together with the `KernelBicop` question, since they're coupled: moving the finite-difference fallback up (both the argument and parameter branches, clamping via the virtual bound getters — each family then overrides or ensures correct clamping) and giving `KernelBicop` a real override (throw for parameter derivatives; exact bilinear gradient for the u1/u2 derivatives with defined edge/corner handling; possibly a spline layer for second derivatives). `scores`/`hessian` are w.r.t. the parameters and already reject nonparametric families, so that path is unaffected.

PR #692, opened 2026-07-11
Speed up Vinecop evaluation and structure selection

tvatter:
Speeds up Vinecop evaluation and structure selection (split out of #681). Independent of the derivatives PR #683: this touches only the evaluation/selection code paths (the `scores`/`hessian` block is untouched, so both merge cleanly in either order).

## Changes

**Evaluation**
- `inverse_rosenblatt` no longer deep-copies a `Bicop` per edge × tree × batch via `as_continuous()` (previously cloned TLL interpolation grids on every access — the dominant cost of simulating from TLL vines).
- `collapse_data_inplace` + cached discrete count + lazy discrete sub-buffers; no-copy fast path for already-collapsed (all-continuous) data.
- Parallelized, allocation-free Monte-Carlo `cdf`.
- `pdf_full(keep_all=true)` no longer fills `hfunc*_sub` with duplicated continuous h-functions (matches the documented discrete-only contract); the test pins both the continuous (empty) and discrete (filled) behavior.

**Selection**
- NaN-free fast path for the tree criterion (no per-candidate data copy); weights/criterion invariants hoisted; h-function storage *moved* between trees; h-functions skipped on the final tree; sparse-selection snapshots stripped of dead data; candidates fitted in place.

## Numerical impact

Bit-identical except `vinecop` log-likelihoods ≤ 1e-12 (documented gates). Full Debug suite (incl. R-parity) passes on current `dev`.

> Review update: the rolling-bitmask rewrite of the proximity check was dropped (see the review thread) — measured, it is 17–77× faster in relative terms but the check runs once per structure construction and costs milliseconds at most, so readability wins. A behavioral large-d proximity test (d=70, pass + violation) was kept.

## Benchmarks

Idle 64-core host, single-threaded (except threads=4 cases), 4 interleaved rounds × 8 repetitions, median-of-medians vs `dev` (Gaussian vines; ratio < 1 = faster):

| case | dev (ns) | PR (ns) | ratio |
|---|---|---|---|
| vinecop/simulate/d=5/n=1000 | 1,398,809 | 1,098,961 | **0.79** |
| vinecop/inverse_rosenblatt/d=5/n=1000 | 1,414,114 | 1,116,423 | **0.79** |
| vinecop/simulate/d=25/n=1000 | 42,978,810 | 34,199,281 | **0.80** |
| vinecop/inverse_rosenblatt/d=25/n=1000 | 44,257,446 | 35,426,469 | **0.80** |
| vinecop/pdf/d=5/n=1000/threads=4 | 1,155,367 | 1,095,092 | 0.95 |
| vinecop/cdf/d=5/n=100/N=10000 | 23,734,859 | 22,723,223 | 0.96 |
| vinecop/select/tll/d=5/n=1000/threads=4 | 135,117,935 | 130,607,542 | 0.97 |
| vinecop/select/itau/d=10/threads={1,4} | — | — | ~1.00 |

The 1.25× on `simulate`/`inverse_rosenblatt` is the per-edge `Bicop`-copy removal on *Gaussian* vines; for TLL vines (where each copy cloned interpolation grids) the effect is much larger — that was the dominant cost of simulating from TLL vines.

### DIFF
--- a/include/vinecopulib/vinecop/implementation/rvine_structure.ipp
+++ b/include/vinecopulib/vinecop/implementation/rvine_structure.ipp
@@ -4,7 +4,6 @@
 // the MIT license. For a copy, see the LICENSE file in the root directory of
 // vinecopulib or https://vinecopulib.github.io/vinecopulib/.
 
-#include <cstdint>
 #include <vinecopulib/misc/tools_serialization.hpp>
 #include <vinecopulib/misc/tools_stats.hpp>
 #include <vinecopulib/misc/tools_stl.hpp>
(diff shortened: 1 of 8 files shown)


include/vinecopulib/vinecop/implementation/rvine_structure.ipp
```diff
@@ -683,21 +684,50 @@ RVineStructure::check_antidiagonal() const
...
-        test_set[i] = struct_array_(i, min_array_(t, e) - 1);
+      const size_t m = min_array_(t, e);
+      const size_t b1 = struct_array_(t, e);
+      const size_t b2 = m;
+      bool equal = true;
+      for (size_t k = 0; k < w; ++k) {
+        uint64_t m1 = masks[e * w + k];
+        uint64_t m2 = masks[(m - 1) * w + k];
+        if ((b1 >> 6) == k) {
+          m1 |= (uint64_t(1) << (b1 & 63));
```
tnagler: This code is close to unreadable. Unless this gives huge performance gains, I would prefer keeping the old version for the structure check

tvatter: Agreed — reverted to the set-based check. For the record, the measured difference (construction with `check=true`, isolated check cost, Tiger Lake): 50µs → 3µs at d=25, 2.7ms → 54µs at d=100, 20.6ms → 0.27ms at d=200. Large relative speedup, but the check runs once per constructed structure and its absolute cost is milliseconds even at d=200, so it's not a workflow-level gain and readability wins. I kept a behavioral large-d test (d=70, valid structure passes + proximity violation throws) for coverage. Also rebased the branch onto current dev.

tvatter: Agreed, change reverted

PR #696, opened 2026-07-14
feat(vinecop): conditional sampling (simulate_conditional)

tvatter:
## What

Adds **conditional sampling** to `Vinecop` — draw a subset of variables from their conditional distribution given fixed values of the rest (matching the Python *VineCopulas* `sample_vinecopconditional`):

```cpp
Eigen::MatrixXd simulate_conditional(
  const Eigen::MatrixXd& u_cond,
  bool qrng = false, size_t num_threads = 1,
  const std::vector<int>& seeds = {}) const;
```

**PR 1 of 2**; conditioning-aware structure *selection* follows in a second PR.

## API

The conditioning variables are the **last `k` of `get_order()`** — they're drawn first by the vine and form a self-contained sub-vine, so any tail is admissible (no separate "conditioning set" argument, and no inadmissible-set error path). Each **row** of `u_cond` is one conditioning point, so `n = u_cond.rows()` (no separate `n` argument). Column `i` holds variable `get_order()[d-k+i]`. To draw many samples at a single conditioning point, pass it repeated (`u0.replicate(n, 1)`).

For **all-continuous** conditioning, `u_cond` is `n × k`. **Discrete** conditioning variables are supported: as with the other data-taking methods, a discrete variable also needs its left-limit CDF `F(x⁻)`, so `u_cond` carries one extra column per discrete conditioning variable — `n × (k + k_d)`, with the `k_d` left-limit columns after the `k` value columns (same layout as the collapsed `n × (d + n_discrete)` input elsewhere). `k` is inferred from the column count.

## How it works

Reuses the existing, tested primitives — no new numerical cascade:
1. embed `u_cond` into a full data matrix in the collapsed `n × (d + n_discrete)` layout (fillers irrelevant — the conditioning block is self-contained);
2. `rosenblatt(..., randomize_discrete = true)` → the independent-uniform seeds of that block (for a discrete conditioning value the seed is drawn across the atom's `[F(x⁻), F(x)]` interval, Brockwell; a no-op for continuous models);
3. splice those seeds into a fresh `simulate_uniform(n, d, qrng, seeds)` matrix;
4. `inverse_rosenblatt(...)` → the free variables are drawn from their conditional distribution; continuous conditioning columns reproduce `u_cond` exactly, discrete ones reproduce the value up to its atom.

## Discrete support

> **Update (addressing @tnagler's review).** The first revision threw on discrete variables; as suggested, they are now handled via `rosenblatt`'s `randomize_discrete` option. Continuous behaviour is bit-for-bit unchanged (`randomize_discrete` is a no-op without discrete variables — the continuous correctness test is unaffected). The conditioned (free) draws are correct with multiple discrete conditioning variables too; only the *echoed* conditioning column of a later-drawn discrete variable may land slightly outside its atom (documented on the method).

## Tests (extend `test_vinecop_class`)

- conditioning columns reproduced exactly for continuous vars (fixed-point via `.replicate` **and** per-sample), `all_close(1e-9)`; multi-thread == single-thread (`1e-10`); seed determinism.
- **statistical correctness** (continuous): conditional mean of a free variable matches a large unconditional sample filtered near the conditioning values, and clearly differs from the unconditional mean.
- **discrete** (`simulate_conditional_discrete_is_correct`): conditioning column lands in its atom `[F(x⁻), F(x)]`; conditional mean matches the unconditional sample filtered to the atom interval; determinism.
- **multi-discrete** (`simulate_conditional_multi_discrete`): two stacked discrete conditioning variables — first-drawn column reproduced within its atom, and the free-variable conditional mean matches the atom-filtered reference (the empirical bias is ~0.002 vs a ~0.15 conditioning shift).
- error paths: 0 columns, column count matching no `k`, out-of-cube values, `F(x⁻) > F(x)`.

Verified under **Debug + ASan/UBSan** and **precompiled release** (`VINECOPULIB_PRECOMPILED=ON`); `clang-format-14` clean; existing `simulate`/`rosenblatt` tests unchanged.

## API stability

Purely additive (one new method) — no breaks. Header-only, no ABI concern. NEWS.md updated under `### NEW FEATURES`.


include/vinecopulib/vinecop/implementation/class.ipp
```diff
@@ -2292,6 +2292,86 @@ Vinecop::simulate(const size_t n,
...
+//!
+//! @details The conditioning variables are the last `k = u_cond.cols()`
+//! variables of the vine order `get_order()` (they are drawn first by the vine
+//! and form a self-contained sub-vine). Each row of `u_cond` is one
+//! conditioning point; the corresponding output row is drawn from the
+//! remaining variables' distribution conditional on that point. It is
+//! implemented as a Rosenblatt transform of the conditioning variables
+//! followed by an inverse Rosenblatt transform (see `rosenblatt()` /
+//! `inverse_rosenblatt()`); the conditioning columns of the result therefore
+//! reproduce `u_cond` exactly. Only continuous variables are supported.
```
tnagler: Should be fairly easy to allow discrete variables with the randomize_discrete option in rosenblatt() https://github.com/vinecopulib/vinecopulib/blob/6672ccf9e16f1b14130ea55826e45daeba837e86/include/vinecopulib/vinecop/implementation/class.ipp#L1253

tvatter: Indeed, done.

PR #699, opened 2026-07-20
feat(bicop,vinecop): scores/gradient/Hessian on Bicop and parameter-vectorized scores/Hessian

tvatter:
## Summary

Fills two gaps left by the recent derivative work (#675/#681/#683/#687):

1. **`Bicop` scores/gradient/Hessian.** The per-observation analytic derivatives existed, but there was no aggregated score/gradient/Hessian API on the bivariate facade — only on `Vinecop`. Adds `scores` (n×p), `gradient` (p), `hessian` / `hessian_full` (averaged and per-observation p×p), `scores_cov`, and `scores_full` as thin aggregations of `logpdf_deriv`/`logpdf_deriv2`, mirroring the `Vinecop` asymptotic API. Each also has a **per-row-parameter** overload. Parametric families with continuous variables only (inherits the existing derivative preconditions).

2. **Parameter-vectorized scores/Hessian.** Adds per-observation-parameter overloads of `Vinecop::scores`, `scores_full`, `gradient`, `hessian`, `hessian_full`, and `scores_cov`, taking an `n × npars` matrix (one full-vine parameter vector per observation, columns in the `(tree, edge, parameter)` order of `scores()`). Row `i` is evaluated as if the vine carried that row's parameters — enabling scores/Hessians for conditional/covariate vine models where the pair-copula parameters vary by observation. Continuous, all-parametric models only (discrete variables and nonparametric pair copulas are rejected).

## No regression to the fixed-parameter path

The shared forward walk `build_deriv_cache` gains an optional per-observation parameter matrix. When supplied, each edge's density, h-function, **and** derivative evaluations switch to the matching per-row `Bicop` overloads; when empty (the default), the existing scalar path runs unchanged, so the 99% of users who don't need per-observation parameters pay nothing. All downstream cascades (scores, both Hessian variants, `scores_cov`) and the `DerivCache`/`ScoresResult` data structures are reused unchanged.

## Tests

- **Bicop:** scores/gradient/Hessian vs finite differences and vs a per-row loop over single-parameter copulas, for all parametric families × rotations; threading parity, broadcast reduction, discrete/nonparametric rejection.
- **Vinecop:** per-observation `scores` and `hessian_full` checked row-by-row against a vine rebuilt with that observation's parameters (both `step_wise` settings); fixed-path parity when a single parameter vector is broadcast; thread-count determinism; discrete/shape rejection. All pre-existing derivative tests continue to pass unchanged.

Locally: `clang-format-14` clean; Debug (ASAN) build; 97 Bicop + 11 Vinecop targeted tests pass (the R-oracle parity test is left to CI).

### DIFF
--- a/include/vinecopulib/bicop/implementation/class.ipp
+++ b/include/vinecopulib/bicop/implementation/class.ipp
@@ -1060,6 +1060,231 @@ Bicop::logpdf_deriv2(const Eigen::MatrixXd& u,
 }
 //! @}
 
+//! @name Scores, gradient, and Hessian of the log-likelihood
+//!
+//! @details These methods aggregate the log-density parameter derivatives into
+//! the score (the gradient of each observation's log-density contribution), its
+//! observation-average (`gradient`), and the Hessian. They are thin wrappers
+//! around `logpdf_deriv()` / `logpdf_deriv2()`: with `p = get_parameters()`
+//! parameters, the score of observation `i` w.r.t. parameter `k` is
+//! \f$ \partial \log c(u_i; \theta) / \partial \theta_k \f$, and the (per-obs
+//! and averaged) Hessians collect the second log-density derivatives. Like the
+//! derivatives they build on, they require parametric families and continuous
+//! variable types; nonparametric or discrete models throw.
+//!
+//! The per-row-parameter overloads evaluate at a different parameter set per
+//! row of `u` (see the corresponding `pdf()` overload for the layout and
+//! validation rules).
+//!
+//! @param u An \f$ n \times 2 \f$ matrix of observations contained in
+//!   \f$ (0, 1)^2 \f$.
+//! @{
+
+//! @brief Evaluates the per-observation scores.
+//!
+//! @return An \f$ n \times p \f$ matrix whose column `k` is
+//! \f$ \partial \log c / \partial \theta_{k+1} \f$.
+inline Eigen::MatrixXd
+Bicop::scores(const Eigen::MatrixXd& u) const
+{
+  check_data(u);
+  check_deriv_preconditions();
+  const Eigen::Index p = static_cast<Eigen::Index>(deriv_npars());
+  Eigen::MatrixXd s(u.rows(), p);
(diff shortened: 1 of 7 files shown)


include/vinecopulib/bicop/implementation/class.ipp
```diff
@@ -1060,6 +1060,231 @@ Bicop::logpdf_deriv2(const Eigen::MatrixXd& u,
...
+//! @param u An \f$ n \times 2 \f$ matrix of observations contained in
+//!   \f$ (0, 1)^2 \f$.
+//! @{
+
+//! @brief Evaluates the per-observation scores.
+//!
+//! @return An \f$ n \times p \f$ matrix whose column `k` is
+//! \f$ \partial \log c / \partial \theta_{k+1} \f$.
+inline Eigen::MatrixXd
+Bicop::scores(const Eigen::MatrixXd& u) const
```
tnagler, line 1143: lots of duplicated code here. shouldn't the simpler versions just call the parametrized versions?

tvatter: Done in 27b4c60 — the shared loop now lives once.

I kept them as separate overloads rather than having the simple ones call the parametrized ones, because that particular delegation would actually cost us on the hot path:

- the parametrized overload takes an `n × p` matrix, so a simple call would have to `replicate()` the stored parameters to every row,
- then pay `format_parameters` (per-column bounds + `allFinite`) validation, and
- most importantly, lose the single-row broadcast: the internal leaves special-case `parameters.rows() == 1` to compute the parameter-dependent constants once (e.g. the Student df-only terms hoisted in #693); an `n × p` matrix forces that per row.

So instead I factored the duplicated *loop* into three helpers (`assemble_scores` / `assemble_hessian` / `assemble_hessian_full`, in the spirit of vinecop's `propagate_first_order`) that take a per-column derivative evaluator. Each overload passes its own closure — the simple one keeps the broadcast `logpdf_deriv` path, the per-row one keeps the `n × p` path — so the scaffolding is written once with no delegation and no behavior change. Scores/Hessian tests still pass.

PR #701, opened 2026-07-23
perf(vinecop): speed up the RVineTrees peel (byte-identical)

tvatter:
## What

`RVineTrees` (the shared leaf-peeling primitive behind `Vinecop::reorient()` and `select()`'s `finalize`, added in #698) reorganized the hand-rolled peel but ran **~3.6× slower** — purely from data-structure choices out of step with the rest of the vine code (which uses `std::vector<size_t>` + `tools_stl`, never `std::set`). This optimizes the internals **with byte-for-byte identical output** — no algorithm change.

## Changes

- **`Edge::C`**: `std::set<size_t>` → **sorted `std::vector<size_t>`** (the codebase idiom), plus a precomputed sorted **`all_indices`** so the descent no longer rebuilds a set per candidate.
- **`AugmentedTree`**: each augmented edge now holds a **non-owning `const Edge*`** instead of a copy of the whole `Edge` — a `Bicop` copy is a *deep* copy (`AbstractBicop::create` + parameter-matrix copy), so this removes **~d²/2 deep `Bicop` copies per peel** (the dominant cost). `degrees` is a `std::vector` indexed by node id instead of a `std::map`.
- **`peel`**: edges are marked `consumed` with a flag instead of `erase()` — preserving iteration order, so the leaf choice, descent match, and flips are unchanged; `check_set` is a sorted vector.
- **`build_lookup`**: sorted-vector keys instead of `std::set` keys.
- **`finalize_unknown_structure`** builds its conditioning as a `std::vector`.

## Performance (`reorient`, D-vine, isolated, median ms)

| d | before (`std::set`) | after | speedup |
|--:|--:|--:|--:|
| 5  | 0.13 | 0.09 | 1.6× |
| 10 | 0.79 | 0.45 | 1.8× |
| 20 | 1.99 | 0.99 | 2.0× |
| 30 | 5.45 | 2.24 | 2.4× |
| 50 | **23.8** | **7.2** | **3.3×** |

Back to ~pre-refactor (hand-rolled) parity. `select()`'s `finalize` benefits proportionally to its (small) share of a fit.

## Byte-identical — verified

The `select()` default path routes through the peel, so output must not move. `std::set` ⟺ sorted `std::vector` (same canonical order), `consumed`-flag ⟺ `erase` (same survivor order → same leaf/descent choices), pointer-vs-copy doesn't change values, precomputed `all_indices` reproduces `set(C)∪{a,b}` exactly.

- **Golden harness** (28 configs: all tree criteria/algorithms incl. random+seeds, sparse/threshold, truncated, discrete, weights): `select()` structure + hexfloat loglik/mBICv/params **dump *and* trace byte-identical** before/after.
- `rvine_trees_works` (round-trip `(order, struct_array)` identity + `RVineStructure` round-trips), `rvine_trees_sanity_checks` (throws), `reorient_preserves_model` (pdf 1e-6, tail==B), `conditional_select_empty_equals_plain` (**exact `get_matrix()` equality** vs plain select), `conditional_select_places_conditioning_set_at_tail`, `simulate_conditional_discrete_is_correct` — all pass.
- Debug `STRICT_COMPILER=ON` build clean; `clang-format-14` clean. Full `test_all` (precompiled, sanitizers, R-parity) left to CI.

No `NEWS.md` change — behaviour is identical (#698's bullet already says "byte-for-byte unchanged").

### DIFF
--- a/include/vinecopulib/misc/tools_stl.hpp
+++ b/include/vinecopulib/misc/tools_stl.hpp
@@ -51,6 +51,17 @@ intersect(std::vector<T> x, std::vector<T> y)
   return common;
 }
 
+//! @brief Inserts `x` into the sorted vector `v`, keeping it sorted and unique
+//! (mirrors `std::set::insert`).
+template<class T>
+void
+insert_sorted(std::vector<T>& v, const T& x)
+{
+  auto it = std::lower_bound(v.begin(), v.end(), x);
+  if (it == v.end() || *it != x)
+    v.insert(it, x);
+}
+
 template<class T>
 size_t
 find_position(T x, const std::vector<T>& vec)
--- a/include/vinecopulib/vinecop/implementation/rvine_trees.ipp
+++ b/include/vinecopulib/vinecop/implementation/rvine_trees.ipp
@@ -21,9 +21,10 @@ inline RVineTrees::RVineTrees(const std::vector<size_t>& order,
   trees_.resize(trunc_lvl_);
   for (size_t t = 0; t < trunc_lvl_; ++t) {
     for (size_t e = 0; e < d_ - 1 - t; ++e) {
-      std::set<size_t> conditioning;
+      std::vector<size_t> conditioning;
+      conditioning.reserve(t);
       for (size_t k = 0; k < t; ++k)
-        conditioning.insert(struct_array(k, e));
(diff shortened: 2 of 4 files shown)


tnagler:
LGTM

PR #735, opened 2026-08-17
fix(vinecop): report select_threshold on the "Select threshold" line

tvatter:
## What

`FitControlsVinecop::str()` printed `get_select_trunc_lvl()` under **both** the
"Select truncation level" and the "Select threshold" label:

```cpp
controls_str << "Select threshold: "
             << static_cast<std::string>(get_select_trunc_lvl() ? "yes" : "no")
```

The two lines are adjacent and otherwise identical, so the wrong getter still
prints a plausible yes/no and nothing looks amiss. But whenever the two flags
differ, anyone printing their controls to check a sparse-selection setup is
told the opposite of what they configured:

```cpp
FitControlsVinecop c;
c.set_select_threshold(true);   // c.str() reports "Select threshold: no"
```

## Also in this PR

The declaration of `tools_stats::dependence_matrix` is removed. It appears
exactly once in the whole library — a declaration in `misc/tools_stats.hpp`
with no definition and no caller — so it is not part of the API in any usable
sense: naming it from a translation unit produces an undefined symbol at link
time. Found while trying to bind it for Python, where it linked a wheel that
failed to import.

## Test

`vinecop_sanity_checks.controls_print` previously only asserted that `str()`
does not throw, which a copy-paste like this passes. It now sets the two flags
against each other and reads both lines back, so each has to report itself.
It fails on `main` and passes here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -146,6 +146,12 @@ wrong thing.
 
 ### BUG FIXES
 
+* Report `select_threshold` rather than `select_trunc_lvl` on the
+  "Select threshold" line of `FitControlsVinecop::str()` (#735)
+
+* Remove the declaration of `tools_stats::dependence_matrix`, which had no
+  definition anywhere in the library, so any call to it failed to link (#735)
+
 * Treat omitted pair-copulas as implicit independence during evaluation of a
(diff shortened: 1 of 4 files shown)


tnagler (approved):
LGTM

PR #739, opened 2026-08-20
fix(bicop): compute a real discrete density for kernel pair copulas

### LINKED ISSUE #736: TllBicop::fit discards find_latent_sample, so the discrete TLL grid equals the continuous one
`TllBicop::fit` computes a latent sample for discrete data and then never uses it, so the fitted TLL grid is the same as it would be for continuous data.

`include/vinecopulib/bicop/implementation/tll.ipp` (checked at `79e70c3`):

```cpp
  // use jittering in case observations are discrete
  auto psobs =
    tools_stats::to_pseudo_obs(data.leftCols(2), "random", weights, { 5 });
  Eigen::MatrixXd z_data = tools_stats::qnorm(psobs);          // (1)

  Eigen::Matrix2d B = select_bandwidth(z_data, method, weights);
  B *= mult;

  // find latent sample in case observations are discrete
  if (var_types_[0] == "d" || var_types_[1] == "d") {
    psobs =
      tools_stats::find_latent_sample(data, std::pow(B(0, 0) * B(1, 1), 0.25));   // (2)
  }

  Eigen::MatrixXd ll_fit = fit_local_likelihood(z, z_data, B, method, weights);   // (3)
```

`z_data` is computed at (1) from the jittered ranks and is what (3) fits on. The
latent sample assigned at (2) is never read: `psobs` is next used only in the
discrete EDF branch further down, which immediately overwrites it with mid ranks

```cpp
    psobs = 0.5 * (data.leftCols(2) + data.rightCols(2)).array();
```

so `find_latent_sample`'s result is dead. The comment at (2) and the existence of
the function suggest the intent was `z_data = tools_stats::qnorm(psobs);` after
the branch.

### Reproduction

Same four-column data, fitted twice with different `var_types`:

```python
import numpy as np, pyvinecopulib as pv

rng = np.random.default_rng(0)
n = 2000
z = rng.standard_normal((n, 2))
z[:, 1] = 0.7 * z[:, 0] + 0.7 * rng.standard_normal(n)
pmf = np.array([1.0, 4.0, 6.0, 4.0, 1.0]) / 16.0
cdf = np.cumsum(pmf)
k = np.searchsorted(cdf, pv.to_pseudo_obs(z[:, [0]]).ravel())
u1, u1m = cdf[k], np.where(k > 0, cdf[np.maximum(k - 1, 0)], 0.0)
u2 = pv.to_pseudo_obs(z[:, [1]]).ravel()
u4 = np.column_stack([u1, u2, u1m, u2])
ctl = pv.FitControlsBicop(family_set=[pv.families.tll])

cont = pv.Bicop.from_data(u4, controls=ctl, var_types=["c", "c"])
disc = pv.Bicop.from_data(u4, controls=ctl, var_types=["d", "c"])
print(np.array_equal(cont.parameters, disc.parameters))  # True
print(cont.npars, disc.npars)                            # 197.27..., 49.04...
```

The grids are bit-identical; only `npars_` (and hence `aic` / `bic`) differ,
because that is the one place the discrete branch actually feeds into.

### Why it matters downstream

I hit this porting the TLL fit to a pure-PyTorch backend in pyvinecopulib
(vinecopulib/pyvinecopulib#300). Because the grid turns out to be
`var_types`-independent, the port only had to reproduce the *tie-breaking* in the
ranks, which made it a small change. If this is fixed here, that shortcut stops
being valid and the latent-sample step needs porting too — so I would like to
know which way you intend to go.

Happy to send a PR if you agree the missing line is `z_data = qnorm(psobs)`.

tvatter:
Two defects made every discrete or mixed fit with a nonparametric pair copula
wrong. `tll` is the default family in `FitControlsVinecop`, so this was the
default path for discrete data. Neither the test suite nor pyvinecopulib's caught
it, because every discrete test uses a parametric family and every `tll` test
uses continuous data.

Closes #736.

## 1. Evaluation: the discrete density was never computed

`AbstractBicop::pdf` / `hfunc1` / `hfunc2` are `virtual` and `KernelBicop`
overrode all three:

```cpp
// bicop/implementation/kernel.ipp, before
KernelBicop::pdf(const Eigen::MatrixXd& u)
{
  if (u.cols() == 4) {
    // evaluate jittered density at mid rank for stability
    return pdf_raw((u.leftCols(2) + u.rightCols(2)).array() / 2.0,
                   Eigen::MatrixXd());
  }
  return pdf_raw(u, Eigen::MatrixXd());
}
```

So a four-column argument returned the **continuous** interpolated quantity at
the atom midpoint, and `AbstractBicop::pdf_c_d` / `pdf_d_d` and the discrete
`hfunc*` difference quotients were unreachable for kernel families — in fact
unreachable at all, since `Bicop::pdf(u, parameters)`, the only other entry to
them, rejects nonparametric families in `format_parameters`.

### Why the midpoint rule is a defect and not a modelling choice

A discrete/continuous density must satisfy, for every `u2`,

```
sum_atoms c(u1, u2) * (u1 - u1^-)  =  h2(1, u2) - h2(0, u2)  =  1
```

because the difference quotients telescope over the atoms. Measured on one
fitted 30×30 grid:

| atoms | u2 = 0.1 | u2 = 0.5 | u2 = 0.9 | source |
|---|---|---|---|---|
| 2 | 1.019639 | 1.036184 | 0.899977 | shipped `Bicop::pdf` |
| 2 | **1.000000** | **1.000000** | **1.000000** | difference quotient |
| 8 | 1.013834 | 0.996600 | 1.033555 | shipped `Bicop::pdf` |
| 8 | **1.000000** | **1.000000** | **1.000000** | difference quotient |
| 32 | 1.002292 | 1.001842 | 1.003035 | shipped `Bicop::pdf` |
| 32 | **1.000000** | **1.000000** | **1.000000** | difference quotient |

The quotient is exact at every resolution; the midpoint rule is off by up to
10% and does not converge away. Per cell the error reached **40%**. On a
three-dimensional discrete vine, `Vinecop::pdf` was off by up to a factor of
two and `loglik` was **522.11** against a correct **515.56** — which feeds
`aic` / `bic` / `mbicv` and therefore biases family and structure selection
toward `tll` on discrete data.

All eleven parametric families satisfy the identity to ~1e-15, so `tll`'s
discrete surface disagreed with every parametric one.

### The fix

Deleting the three overrides is sufficient. `KernelBicop::cdf` and the `*_raw`
members already supply everything the inherited dispatchers need, and they ignore
their `parameters` argument, so nothing has to be plumbed. #700 had already
hardened `pdf_c_d` / `pdf_d_d` against the 30×30 grid parameter shape — which is
precisely what makes reaching them safe now.

After the change, `Bicop::pdf` on a discrete kernel edge equals the difference
quotient of its own `hfunc2` **exactly** (max relative difference `0.0` over
3000 rows), and the atom masses sum to `1.0000000000`.

## 2. Fitting: the latent sample was discarded (#736)

`TllBicop::fit` computes a latent sample for discrete data and then throws it
away — `z_data` is built from the jittered ranks at line 247, `find_latent_sample`
assigns to `psobs` at 255, and `fit_local_likelihood` reads `z_data`. `psobs` is
next touched where the EDF branch overwrites it with mid ranks, so the latent
sample has no effect whatsoever. One line inside the existing branch fixes it:

```cpp
  if (var_types_[0] == "d" || var_types_[1] == "d") {
    psobs =
      tools_stats::find_latent_sample(data, std::pow(B(0, 0) * B(1, 1), 0.25));
    z_data = tools_stats::qnorm(psobs);          // added
  }
```

Nothing between the two points reads `psobs`, so this is safe by inspection.

The effect is easiest to see as an identity that used to hold and should not:
fitting the same data as `{"d","c"}`, `{"c","d"}` and `{"d","d"}` returned
**bit-identical grids and bit-identical Kendall's tau**. Recovered tau against a
true 0.6, zero-inflated margins:

| n | before (all three var_types) | after: d/c | after: c/d | after: d/d |
|---|---|---|---|---|
| 20 | 0.5052 | 0.0351 | 0.1393 | 0.1304 |
| 200 | 0.5593 | 0.3361 | 0.3360 | 0.2975 |
| 1000 | 0.5693 | 0.5391 | 0.5551 | 0.5331 |
| 5000 | 0.5759 | 0.5848 | 0.5878 | 0.5825 |

The "before" column is identical across var_types by construction — that is the
bug. Afterwards the estimator actually uses the atom structure and converges to
0.6; the price is genuine small-sample variance, which the old code hid by
ignoring discreteness altogether.

## This is a behavior change

Every discrete or mixed fit with a `tll` pair copula now has a different grid,
density, `loglik`, `aic`, `bic` and `mbicv`, so family and structure selection can
change. **Continuous fits are untouched** — neither code path runs when every
variable type is `"c"`, which is the useful control.

One existing test moved with it: the three `tll` fits in `zero_inflated.bicop`
now use the whole sample instead of `topRows(20)`. Their tau assertions passed
only because the fit ignored the atoms; once the latent sample reaches it, 20
observations no longer identify a nonparametric density. **The tolerance is
unchanged** — only the sample size, which was arbitrary.

## Tests

- `discrete.kernel_pdf_is_normalized` — the identity above for `tll` at 2, 8 and
  32 atoms and three `u2`, with `gaussian` as a control. It needs no reference
  implementation and no tolerance argument.
- `discrete.kernel_fit_uses_the_latent_sample` — a discrete fit must no longer
  reproduce the grid the same data produces when declared continuous.

Full suite: **714 tests pass**.

## Two follow-ups, deliberately not in this PR

- `pdf_c_d` / `pdf_d_d` and the discrete `hfunc*` branches loop row by row,
  constructing an `Eigen::MatrixXd par_i` per observation. For kernels that is
  now on the hot path, so a vectorized `KernelBicop::pdf_c_d` / `pdf_d_d`
  override is worth having. This PR is the correctness half only.
- With the overrides gone nothing overrides `pdf` / `hfunc1` / `hfunc2` any more,
  so they need not stay `virtual`. Left alone here to keep the diff to the
  defect.

## Related, in wdm

`rank(..., "random")` accumulates the tie weight in the wrong order
(`wdm/include/wdm/ranks.hpp`, the `"random"` branch adds `ww` before
accumulating it where `"first"` accumulates first). A tie group of 6 occupying
rank slots 4..9 comes back as `[4,4,5,6,7,8]` — minimum duplicated, maximum
dropped — so `to_pseudo_obs(x, "random")` returns data that still contains ties
and is not idempotent. Its only call site here is `tll.ipp`'s jittering, which
after this PR only seeds the bandwidth. Filed separately.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -64,6 +64,12 @@ wrong thing.
 
 ### BEHAVIOR CHANGES
 
+* Every discrete or mixed fit with a nonparametric (`tll`) pair copula changes:
+  the fitted grid, density, `loglik`, `aic`, `bic` and `mbicv` all move, so
+  family and structure selection can differ. `tll` is the default family, so
+  this is the default path for discrete data. Continuous fits are unaffected;
+  see BUG FIXES (#739)
+
 * TLL fits change near the boundary of the unit square: the endpoint clamping of
   the interpolation grid was removed, so every `tll` fit differs slightly from
   0.7.3 (#654)
@@ -145,6 +151,19 @@ wrong thing.
 * Exit structure selection early when the graph is already a tree (#661)
 
 ### BUG FIXES
+
+* Compute a real discrete density for kernel pair copulas. `KernelBicop`
+  overrode `pdf`, `hfunc1` and `hfunc2` to return the continuous quantity at the
+  atom midpoint, so the difference quotients in `AbstractBicop::pdf_c_d` /
+  `pdf_d_d` were unreachable. The midpoint rule violates
+  `sum_atoms c * (u1 - u1^-) = 1` by up to 10% without converging, reached 40%
+  per cell, and shifted a three-dimensional discrete vine's `loglik` by 6.5
+  (#739)
+
+* Use the latent sample when fitting a `tll` pair copula to discrete data.
+  `TllBicop::fit` assigned `find_latent_sample`'s result to `psobs` but fitted on
+  `z_data`, which was already built from the jittered ranks, so declaring a
+  variable discrete had no effect on the fitted grid at all (#736, #739)
 
 * Report `select_threshold` rather than `select_trunc_lvl` on the
   "Select threshold" line of `FitControlsVinecop::str()` (#735)
--- a/include/vinecopulib/bicop/abstract.hpp
+++ b/include/vinecopulib/bicop/abstract.hpp
@@ -100,7 +100,8 @@ protected:
 
   virtual void flip() = 0;
 
-  // following are virtual so they can be overriden by KernelBicop
+  // state-based dispatchers: they read var_types_ and route discrete
+  // arguments through the difference quotients below
   virtual Eigen::VectorXd pdf(const Eigen::MatrixXd& u);
 
   virtual Eigen::VectorXd hfunc1(const Eigen::MatrixXd& u);
--- a/include/vinecopulib/bicop/implementation/kernel.ipp
+++ b/include/vinecopulib/bicop/implementation/kernel.ipp
@@ -31,17 +31,6 @@ KernelBicop::pdf_raw(const Eigen::MatrixXd& u, const Eigen::MatrixXd&)
 }
 
 inline Eigen::VectorXd
-KernelBicop::pdf(const Eigen::MatrixXd& u)
-{
-  if (u.cols() == 4) {
-    // evaluate jittered density at mid rank for stability
-    return pdf_raw((u.leftCols(2) + u.rightCols(2)).array() / 2.0,
-                   Eigen::MatrixXd());
-  }
-  return pdf_raw(u, Eigen::MatrixXd());
-}
-
-inline Eigen::VectorXd
 KernelBicop::cdf(const Eigen::MatrixXd& u, const Eigen::MatrixXd&)
 {
   return interp_grid_->integrate_2d(u);
@@ -57,28 +46,6 @@ inline Eigen::VectorXd
 KernelBicop::hfunc2_raw(const Eigen::MatrixXd& u, const Eigen::MatrixXd&)
 {
   return interp_grid_->integrate_1d(u, 2);
-}
-
-inline Eigen::VectorXd
-KernelBicop::hfunc1(const Eigen::MatrixXd& u)
-{
-  if (u.cols() == 4) {
-    auto u_avg = u;
-    u_avg.col(0) = (u.col(0) + u.col(2)).array() / 2.0;
-    return hfunc1_raw(u_avg.leftCols(2), Eigen::MatrixXd());
-  }
-  return hfunc1_raw(u, Eigen::MatrixXd());
-}
-
-inline Eigen::VectorXd
-KernelBicop::hfunc2(const Eigen::MatrixXd& u)
-{
-  if (u.cols() == 4) {
-    auto u_avg = u;
-    u_avg.col(1) = (u.col(1) + u.col(3)).array() / 2.0;
-    return hfunc2_raw(u_avg.leftCols(2), Eigen::MatrixXd());
-  }
-  return hfunc2_raw(u, Eigen::MatrixXd());
 }
 
 inline Eigen::VectorXd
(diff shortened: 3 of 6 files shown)


tnagler (approved):
good catch! LGTM

PR #743, opened 2026-08-20
fix(vinecop): read an empty pair-copula store as independence everywhere

### LINKED ISSUE #737: Vinecop::reorient reads out of bounds when the vine has no pair copulas
`Vinecop::reorient` indexes `pair_copulas_` without the guard `get_pair_copula` has, so calling it on a vine constructed without pair copulas reads out of bounds.

The constructor documents an empty `pair_copulas` as "all pair-copulas are
treated as independence" but leaves the member empty rather than filling it
(`class.ipp`, checked at `79e70c3`):

```cpp
  if (!pair_copulas.empty()) {
    set_all_pair_copulas(pair_copulas);
  }
```

`reorient` then does, unconditionally:

```cpp
  for (size_t tree = 0; tree < d_ - 1; ++tree) {
    pair_copulas[tree].reserve(d_ - 1 - tree);
    for (size_t edge = 0; edge < d_ - 1 - tree; ++edge) {
      auto location = reorientation.pair_copulas(tree, edge);
      pair_copulas[tree].emplace_back(
        pair_copulas_[location.tree][location.edge]);      // <-- no bounds check
```

so `Vinecop(structure).reorient({1})` dereferences `pair_copulas_[0]` on an empty
vector. `get_pair_copula` guards the same access (`tree >= pair_copulas_.size()`
throws), and `set_var_types_internal` early-returns on `pair_copulas_.empty()`,
so `reorient` looks like the outlier rather than the intended behavior.

Note the `identity` fast path returns before the loop, so a set that already is
the order tail happens to be safe — which makes this easy to miss.

### Where it bites

I wanted the *structure* half of a reorientation from Python, with no fitted
model involved — the natural spelling is a throwaway independence vine on the
same structure, `Vinecop.from_structure(structure=s).reorient(cs)`. That is UB
today. The workaround is to pass a full store of independence copulas
explicitly:

```python
indep = pv.Bicop()
placeholder = pv.Vinecop.from_structure(
    structure=s,
    pair_copulas=[[indep] * (d - 1 - t) for t in range(d - 1)],
)
placeholder.reorient(cs)
```

Two possible fixes, either is fine from my side: treat an empty
`pair_copulas_` as independence in `reorient` (mirroring the constructor's
documented meaning), or have it throw the same message `get_pair_copula` does.

Happy to send a PR.

tvatter:
Closes #737.

`Vinecop`'s constructor documents an empty `pair_copulas` as "all pair-copulas are treated as independence" and leaves the member empty. #729 taught the evaluation paths that convention through `get_effective_trunc_lvl()`. Four paths still assumed a full store — the reported one and three more found while fixing it. Each was confirmed failing under `-DVINECOPULIB_SANITIZERS=ON` before being touched.

| path | before |
| --- | --- |
| `reorient()` | `pair_copulas_[location.tree][location.edge]` unconditionally → `SEGV` at `class.ipp:573` on `Vinecop(structure).reorient({1})` (the reported bug). The `identity` fast path returns before the loop, which is why a set that already was the order tail happened to be safe. |
| `get_trees()` | handed the empty store to `RVineTrees`, which indexes `[t][e]` for every edge the structure has → the same `SEGV`, via `rvine_trees.ipp:57` |
| `truncate()` | resized the empty store to `trunc_lvl`, giving it tree slots holding no pair copula at all; `get_effective_trunc_lvl()` then reported those levels as present and `pdf()` dereferenced them → `SEGV` in `pdf_full` |
| JSON reader | the `try`/`catch` around `pcs_json["tree" + k]` never fired: non-const `operator[]` *inserts* a null instead of throwing. A vine written without pair copulas came back as `Bicop(null)` and threw `json.exception.type_error.305`, so a structure-only vine could not round-trip through `to_file` / `from_file` at all. |

Per @tnagler's vote on the issue, all four treat the empty store as independence rather than throwing.

### The invariant

`pair_copulas_` is either **empty** — every pair copula is independence, whatever the structure's truncation level — or holds **exactly `rvine_structure_.get_trunc_lvl()`** trees, tree `t` having `d - 1 - t` edges. `set_all_pair_copulas()` already maintained it; the reader now does too, truncating the structure to the number of trees a short file actually carries, so the partially-filled state stays unreachable. The reader also reaches its required nodes with `at()`, so a foreign file throws instead of tripping nlohmann's assert and reading past a non-object.

### Tests

- `VinecopConstructor.omitted_pair_copulas_survive_structure_operations` walks a structure-only vine through `reorient`, `get_trees`, `truncate` + `pdf`, and a discrete-`var_types` relabeling. A plain `TEST`, so it runs without Rscript — unlike the `VinecopTest` fixture the other `reorient` tests hang off.
- `serialization.structure_only_vinecop_cbor_roundtrip` covers the reader.

### Verification

- `716/716` with `-DVINECOPULIB_SANITIZERS=ON` (Debug, ASAN + UBSAN, R parity tests included).
- `716/716` with `-DVINECOPULIB_PRECOMPILED=ON` (Release).
- `clang-format-14 --dry-run --Werror` clean on all three files.

NEWS.md entry follows in the next commit, once this PR has a number.

### Noted, not fixed here

`Vinecop(structure).fit(data)` silently does nothing: `fit` early-returns on `get_effective_trunc_lvl() == 0`, so fitting a structure whose families were never set is a no-op and the model only fails later at `check_fitted`. That is a behavior decision with downstream reach (fit independence-initialized families, or throw), so it wants its own issue rather than a drive-by change here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -152,6 +152,12 @@ wrong thing.
 
 ### BUG FIXES
 
+* Read an empty pair-copula store as independence in the paths that still
+  assumed a full one, completing (#729). `Vinecop::reorient` and `get_trees`
+  indexed the store out of bounds, `truncate` gave it tree slots holding no pair
+  copula at all, which `pdf` then dereferenced, and deserializing a vine written
+  without pair copulas threw a JSON type error instead of round-tripping (#743)
+
 * Compute a real discrete density for kernel pair copulas. `KernelBicop`
   overrode `pdf`, `hfunc1` and `hfunc2` to return the continuous quantity at the
   atom midpoint, so the difference quotients in `AbstractBicop::pdf_c_d` /
--- a/include/vinecopulib/vinecop/implementation/class.ipp
+++ b/include/vinecopulib/vinecop/implementation/class.ipp
@@ -134,26 +134,32 @@ inline Vinecop::Vinecop(
 //!      a valid R-vine structure.
 inline Vinecop::Vinecop(const nlohmann::json& input, const bool check)
 {
-  rvine_structure_ = RVineStructure(input["structure"], check);
+  rvine_structure_ = RVineStructure(input.at("structure"), check);
   d_ = static_cast<size_t>(rvine_structure_.get_dim());
   size_t trunc_lvl = rvine_structure_.get_trunc_lvl();
 
-  nlohmann::json pcs_json = input["pair copulas"];
+  const auto pcs_json = input.value("pair copulas", nlohmann::json::object());
   for (size_t tree = 0; tree < std::min(d_, trunc_lvl); ++tree) {
-    nlohmann::json tree_json;
-    try {
-      tree_json = pcs_json["tree" + std::to_string(tree)];
-    } catch (...) {
-      break; // vine was truncated, no more trees to parse
+    const auto tree_key = "tree" + std::to_string(tree);
+    if (!pcs_json.contains(tree_key)) {
+      break; // fewer trees stored than the structure has; the rest are
+             // independence
     }
+    const auto& tree_json = pcs_json[tree_key];
+
     // reserve space for pair copulas of this tree
     pair_copulas_.resize(tree + 1);
     pair_copulas_[tree].resize(d_ - tree - 1);
 
     for (size_t edge = 0; edge < d_ - tree - 1; ++edge) {
-      nlohmann::json pc_json = tree_json["pc" + std::to_string(edge)];
-      pair_copulas_[tree][edge] = Bicop(pc_json);
+      pair_copulas_[tree][edge] =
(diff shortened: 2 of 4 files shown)


tnagler (approved):
All good. 

About:
> Vinecop(structure).fit(data) silently does nothing: fit early-returns on get_effective_trunc_lvl() == 0, so fitting a structure whose families were never set is a no-op and the model only fails later at check_fitted.

I think this is the right behavior, default object corresponds to independence copulas. Replacing them requires model selection, so `select()` must be called.

tvatter:
Follow-ups filed from this work: #746 (`Vinecop::fit` is a silent no-op when the pair-copula store is empty).

PR #744, opened 2026-08-20
fix(bicop): collapse a degenerate atom to its midpoint once in pdf_d_d

### LINKED ISSUE #738: AbstractBicop::pdf_d_d: the degenerate branches read back the midpoint they just wrote
In `AbstractBicop::pdf_d_d`, the two degenerate branches collapse an argument to its atom midpoint with two sequential assignments, and the second reads the value the first just wrote.

`include/vinecopulib/bicop/implementation/abstract.ipp` (checked at `79e70c3`),
the `udiff(i, 0) < 5e-5` branch:

```cpp
      umax(i, 0) = (umax(i, 0) + umin(i, 0)) / 2;
      umin(i, 0) = (umax(i, 0) + umin(i, 0)) / 2;
```

With `a = umax(i, 0)` and `b = umin(i, 0)` on entry, the first line sets
`umax(i, 0) = (a + b) / 2`, and the second then reads that new value, giving
`umin(i, 0) = ((a + b) / 2 + b) / 2 = (a + 3b) / 4` rather than the midpoint
again. The two h-function evaluation points that follow are therefore
`(a + b) / 2` and `(a + 3b) / 4`, differing by `(a - b) / 4`, where the
surrounding code reads as if both should be the midpoint.

The `udiff(i, 1) < 5e-5` branch a few lines down has the same shape on column 1.

Impact is small — the branch only fires when the atom is narrower than `5e-5`,
where the quotient is being replaced by a derivative anyway, so the error is
`O(Δ)` on a quantity that is already an approximation. Reporting it because it is
almost certainly not intended, and because anyone reimplementing `pdf_d_d`
against this as a reference has to decide whether to reproduce it: I did, for
bit-parity, in vinecopulib/pyvinecopulib#299, with a comment pointing here.

Fix is presumably to compute the midpoint once:

```cpp
      const double mid = (umax(i, 0) + umin(i, 0)) / 2;
      umax(i, 0) = mid;
      umin(i, 0) = mid;
```

Happy to send a PR — though it would change `pdf` output in that branch, so it
wants your call on whether it is worth the churn.

tvatter:
Closes #738.

Both degenerate branches of `AbstractBicop::pdf_d_d` assigned the midpoint to one endpoint and then recomputed it from the value they had just written:

```cpp
umax(i, 0) = (umax(i, 0) + umin(i, 0)) / 2;
umin(i, 0) = (umax(i, 0) + umin(i, 0)) / 2;
```

With `a = umax(i, 0)` and `b = umin(i, 0)` on entry, the second line reads the new `umax`, so `umin` became `(a + 3b) / 4`. The two h-function evaluations that follow were `(a - b) / 4` apart in exactly the argument the branch means to have collapsed, so the quotient was not the difference quotient of the other argument alone.

The branch fires only for atoms narrower than `5e-5`, where the quotient is already being replaced by a derivative, so the error is `O(delta)` on a quantity that is already an approximation — but it is not intended, and anyone reimplementing `pdf_d_d` against this as a reference has to decide whether to reproduce it (vinecopulib/pyvinecopulib#299 did, for bit-parity, with a comment pointing at the issue; that workaround can go when the pin moves).

This is the remaining half of the `pdf_d_d` fix in 88825cda (#664), which corrected the `umin(i, 0)` → `umin(i, 1)` index typo in the second branch but left both read-backs in place.

### Behavior change

`pdf` values move for discrete pairs whose atom is narrower than `5e-5`. On the new test's configuration: `1.2196211889514808` → `1.2196363144810516`.

### Tests

`discrete.d_d_degenerate_branch_uses_the_atom_midpoint` pins each branch against the difference quotient of the *continuous* twin's `hfunc1` / `hfunc2` at the midpoint, at `1e-12`. It fails on `main` by `1.5e-5`.

### Verification

- `715/715` Release (header-only) and `715/715` Release with `-DVINECOPULIB_PRECOMPILED=ON`; the discrete area also under ASAN + UBSAN.
- `parity_dump` before/after is **bit-identical** under `scripts/compare_parity.py --strict` (worst relative difference `0.000e+00` in every group), which is expected: the harness has no `{"d", "d"}` case, so it covers `pdf_c_d` but never this branch. No gate relaxation needed. Adding a `{"d", "d"}` case to `benchmarks/src/parity_dump.cpp` would be worth doing separately — a new key has no baseline to compare against, so it does not belong in this PR.
- `clang-format-14 --dry-run --Werror` clean.

NEWS.md entry follows in the next commit, now that this PR has a number.

### Found while verifying, not fixed here

In a `VINECOPULIB_PRECOMPILED=ON` build directory, editing an `.ipp` and running `make` **silently compiles the previous implementation**. The `.ipp` → `.cpp` translation in `cmake/findHeaders.cmake` runs at configure time, and `CONFIGURE_DEPENDS` on the glob only re-triggers CMake when the *set* of `.ipp` files changes, never when one's contents change — so nothing regenerates `<build>/generated/src/.../*.cpp`. This suite caught it (the new test failed against a stale `abstract.cpp`), but it is a trap for anyone iterating in such a build tree, and #711's NEWS entry claims the opposite. CI is unaffected: it configures fresh. Filing separately.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -152,6 +152,12 @@ wrong thing.
 
 ### BUG FIXES
 
+* Collapse a degenerate atom to its midpoint once in `AbstractBicop::pdf_d_d`.
+  Both degenerate branches assigned the midpoint to one endpoint and then
+  recomputed it from the value they had just written, so the second h-function
+  was evaluated at `(a + 3b) / 4` rather than at the midpoint again. `pdf`
+  values change for discrete pairs whose atom is narrower than `5e-5` (#744)
+
 * Compute a real discrete density for kernel pair copulas. `KernelBicop`
   overrode `pdf`, `hfunc1` and `hfunc2` to return the continuous quantity at the
   atom midpoint, so the difference quotients in `AbstractBicop::pdf_c_d` /
--- a/NEWS.md
+++ b/NEWS.md
@@ -64,6 +64,11 @@ wrong thing.
 
 ### BEHAVIOR CHANGES
 
+* The density of a discrete/discrete pair changes where an atom is narrower than
+  `5e-5`: the collapsed argument is now the atom midpoint for both h-function
+  evaluations, so `pdf`, and any `loglik` over such observations, move by
+  `O(atom width)`; see BUG FIXES (#744)
+
 * Every discrete or mixed fit with a nonparametric (`tll`) pair copula changes:
   the fitted grid, density, `loglik`, `aic`, `bic` and `mbicv` all move, so
   family and structure selection can differ. `tll` is the default family, so
--- a/NEWS.md
+++ b/NEWS.md
@@ -156,7 +156,9 @@ wrong thing.
   Both degenerate branches assigned the midpoint to one endpoint and then
   recomputed it from the value they had just written, so the second h-function
   was evaluated at `(a + 3b) / 4` rather than at the midpoint again. `pdf`
-  values change for discrete pairs whose atom is narrower than `5e-5` (#744)
+  values change for discrete pairs whose atom is narrower than `5e-5`. The
+  parity harness now covers the discrete/discrete leaf, which had no case at
+  all, so a change there is no longer invisible to it (#744)
 
 * Compute a real discrete density for kernel pair copulas. `KernelBicop`
   overrode `pdf`, `hfunc1` and `hfunc2` to return the continuous quantity at the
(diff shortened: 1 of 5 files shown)


tvatter:
Follow-ups filed from this work: #745 (an edited `.ipp` leaves a stale generated `.cpp` in a `VINECOPULIB_PRECOMPILED` build tree) and #747 (`parity_dump` has no discrete/discrete case, so this branch is invisible to the parity harness).

tnagler (approved):

Issue #757, opened 2026-08-22
Exact rectangle mass for TllBicop, so a discrete edge need not difference four cdf values


tvatter:
A mixed-discrete density is a rectangle probability divided by the atom's area, so it amplifies **any** absolute error in `cdf` — pure rounding included — by `≈4/(w₁w₂)` in the atom widths. For a kernel (`tll`) pair that error is avoidable: the interpolant is piecewise bilinear, so the exact mass over `[a₁,b₁]×[a₂,b₂]` is a sum of nonnegative hat-function weights against a nonnegative grid, with no cancellation at all.

We implemented that on the Python/torch side as `TorchBicop.rect_mass` and measured both routes against exact rational truth (`fractions.Fraction`, dyadic endpoints so every query point is exactly representable) on a fitted 30×30 `tll` grid:

| route | max relative error |
|---|---|
| four-corner `cdf` difference | 9.2e-15 |
| exact rectangle mass | 1.2e-15 |

7.6× at the `1/8`-wide atoms our parity tests use, and about three orders of magnitude at the widths inner trees reach.

**We cannot use it**, and the reason is why this belongs here rather than downstream. The torch cascade is pinned bit-for-bit against the compiled one. Tree 1's atom widths are themselves differences of tree 0's h-functions, so a 1e-15 pair-level difference between the two implementations arrives at the vine as **8.5e-8 on 6% of rows**. Reformulating one side alone therefore trades a documented parity guarantee for the accuracy — so our discrete edge still differences four `cdf` values, deliberately, to stay identical to `Bicop`'s own quotients.

If `TllBicop` grew the exact routine, both sides could use it and the accuracy would come for free on the C++ evaluator too — `Vinecop`'s discrete densities carry the same amplification today.

Sketch of what it needs:

- `interval_weights(lo, hi) -> w` such that `w · v == ∫_lo^hi f(t) dt` exactly for the piecewise-linear `f` through `(grid_points, v)`, every entry nonnegative (the hat functions are).
- The rectangle mass is then `wx · (V · wy)`.
[...]

PR #770, opened 2026-09-10
fix(vinecop): accumulate the density in log space

### LINKED ISSUE #763: Vinecop::pdf accumulates edge densities linearly, so loglik underflows to -inf
The vine density is built as a running product: the accumulator starts at `1.0` (`vinecop/implementation/class.ipp:1246`) and each edge is multiplied in (`:1326`) — up to `d(d-1)/2` factors. `Vinecop::loglik` is then `pdf(u).array().log().sum()` (`:2994`).

So when the true log-density falls below about `-745`, the product underflows to exactly `0` and the log-likelihood is `-inf`, although the value is perfectly representable in log space.

The information is destroyed inside `pdf()`, so no caller can recover it. A caller can only clip, which substitutes an invented number for a lost one.

**Suggested fix:** accumulate in log space, `sum(log(edge_pdf))`, exposed as a per-observation log-density that `loglik` sums. The per-edge densities already exist — `pdf_full(..., keep_all=true)` returns `pdf_edges`. A genuinely zero edge density still gives `-inf`, which is correct; only the underflow case changes.

Observed at `7d8debec`.

### LINKED ISSUE #773: Vinecop::loglik with per-observation parameters ignores the empty-u convention
Noticed while working on #763 / #770; filing separately rather than widening that diff.

`Vinecop::loglik(u, num_threads)` treats an empty `u` as "report the stored fit":

```cpp
if (u.rows() < 1) {
  return this->get_loglik();
}
```

The per-observation-parameters overload has no such branch, so `loglik(Eigen::MatrixXd(), Eigen::MatrixXd())` sums over zero rows and returns `0` where the other overload returns `get_loglik()`. (With both arguments empty, `per_obs` is false, so it is the fixed-parameter cascade that runs — just without the early return.)

Minor and probably unreachable in practice, since calling the per-observation overload with no observations is not useful. But the two overloads should not disagree about what an empty `u` means. Either add the same early return, or reject an empty `u` there explicitly.

Related: neither `Vinecop::loglik` overload drops `NaN` rows, whereas `Bicop::loglik` does (`AbstractBicop::loglik` calls `tools_eigen::remove_nans`, and the per-row `Bicop::loglik` skips them in its own loop). A `NaN` observation therefore makes the whole vine log-likelihood `NaN` but only reduces the bivariate one. That is arguably the correct behavior for a vine — a row with a missing value has no likelihood — but it is an undocumented asymmetry between the two classes, and worth settling one way or the other.

tvatter:
Closes #763. Also closes #773.

## The problem

The vine density was built as a running product: the accumulator started at `1.0` and each of the up to `d(d-1)/2` edge densities was multiplied in. `Vinecop::loglik()` was then `pdf(u).array().log().sum()`, so once the true log-density fell below about `-745` the product underflowed to exactly `0` and the log-likelihood was `-inf`, although the value is perfectly representable in log space. `aic()`, `bic()` and `mbicv()` all route through `loglik()` and went with it.

The information was destroyed inside `pdf()`, so no caller could recover it — only clip, substituting an invented number for a lost one.

## The fix

`pdf_full()` now keeps a single accumulator in log space and reports it as `logpdf`, with `pdf` its exponential. `Vinecop::logpdf()` exposes the per-observation log-density (plus a per-observation-parameters overload), and `loglik()` sums it.

No clamp is needed. `AbstractBicop::pdf()` already trims every edge density to `[DBL_MIN, DBL_MAX]`, so each term is finite unless the edge density is `NaN`, and `NaN` still propagates. A genuinely zero edge density still gives `-inf`, which is correct; only the underflow case changes.

## Workarounds removed

The underflow had accumulated six clamps, and they were not harmless:

- `scores_full()`'s discrete fallback differenced `pdf(u).array().max(1e-20).log()` twice. On an underflowing row both legs floored to `log(1e-20)`, so the finite difference was silently `0` and the score was **wrong**, not merely imprecise.
- two per-edge `max(1e-20)` clamps and three test-side `max(1e-300)` clamps could not fire at all past the `DBL_MIN` trim, and only read as a hazard.

## Behavior change

`pdf()` is now the exponential of a log-space sum, so its values move at the ~1e-15 level. `scripts/compare_parity.py` reports a worst relative difference of **1.4e-15** on the `vinecop` group, inside the existing `1e-12` gate — no gate relaxation, and every other group is bit-identical:

```
worst relative difference per group:
  bicop_eval     0.000e+00
  bicop_fit      0.000e+00
  tau            0.000e+00
  tll            0.000e+00
  tools_stats    0.000e+00
  vinecop        1.431e-15
all parity gates passed
```

Downstream bit-for-bit pins — pyvinecopulib's torch cascade in particular (see #757) — will need re-baselining.

`vinecop/pdf` benchmarks are unchanged within run-to-run noise at `d = 5`, `10` and `25` (the extra `log` per edge is lost against the per-edge `Bicop::pdf` work).

## Tests

- `VinecopLogPdf.survives_a_density_that_underflows` — a 50-dimensional D-vine truncated at the first tree, all pair copulas Gaussian with `rho = 0.9`, evaluated at a countermonotone row. `pdf()` is exactly `0`; `logpdf()`, `loglik()` and `aic()` match the analytic reference (`49 * log c`, about `-2347`) to `1e-9`.
- `VinecopLogPdf.agrees_with_the_per_edge_densities` — `logpdf` equals the sum of the per-edge log-densities `pdf_full(keep_all = true)` reports, and `pdf` equals its exponential.
- `VinecopTest.pdf_is_correct` gains `logpdf` / `loglik` assertions against the R oracle; `works_multi_threaded` gains a `logpdf` threading-determinism check; `hfuncs_is_correct` pins the new field's shape.

Full suite green (739 tests, Debug, header-only, with `Rscript` present so the R parity fixtures ran rather than skipped).

## Also here: #773, the `loglik` conventions

`Vinecop::loglik()` now leaves out the observations that have no likelihood rather than returning `NaN` for the whole sample, and both overloads honor the empty-`u` convention — the per-observation one used to sum over zero rows and return `0` where the other returns `get_loglik()`.

The NaN split follows what the codebase already does: **evaluation propagates `NaN`** (`unaryExpr_or_nan` / `binaryExpr_or_nan`, and `pdf()` / `logpdf()` with them), **likelihood summation drops it** (`AbstractBicop::loglik` calls `remove_nans`; `Bicop::loglik(u, parameters)` skips them in its own loop). `Vinecop::loglik` was the odd one out. If you would rather it keep propagating, it is a one-line revert of `sum_loglik`.

This belongs in the same PR as the log-space accumulator because the two conventions have to be told apart, and the new test does exactly that: a `NaN` row is a missing value and drops out, while a `-inf` log-density is the model ruling an observation out and is kept. The underflow case from #763 is the third state — `pdf()` is `0`, `log(pdf())` is `-inf`, and `logpdf()` is a finite `-2347`.

`pdf()` is untouched by this part, and the parity sweep confirms `vinecop` is unaffected by it: `remove_nans` does not reorder a NaN-free sample.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -64,6 +64,11 @@ wrong thing.
 
 ### BEHAVIOR CHANGES
 
+* `Vinecop::pdf()` is the exponential of a log-space sum rather than a running
+  product of edge densities, so its values move at the `1e-15` level (observed
+  maximum `1.4e-15` relative across the parity sweep). Downstream bit-for-bit
+  comparisons against this library need re-baselining; see BUG FIXES (#770)
+
 * `to_pseudo_obs(..., "random")` returns different values for a given seed, and
   every discrete `tll` fit moves with it, because the jitter it applies goes
   through that path. wdm's random tie-breaking assigned its offsets one step out
@@ -111,6 +116,12 @@ wrong thing.
   differ for the same model. Densities and log-likelihoods do not (#702)
 
 ### NEW FEATURES
+
+* Add `Vinecop::logpdf()`, the per-observation log-density, alongside a `logpdf`
+  field on the `pdf_full()` result. A vine density is a product of one factor
+  per edge, so `pdf()` underflows to `0` in high dimensions or under strong
+  dependence while the log-density is still an ordinary double; `logpdf()` is
+  the accurate way to obtain it, and `loglik()` sums it (#770)
 
 * Add `"cxi"` as a `tree_criterion`, weighting edges by Chatterjee's xi
   symmetrized as `max(xi(X, Y), xi(Y, X))`. Like `"hoeffd"` it picks up
@@ -190,6 +201,16 @@ wrong thing.
 * Exit structure selection early when the graph is already a tree (#661)
 
 ### BUG FIXES
+
+* `Vinecop::loglik()` is finite whenever the log-likelihood is representable,
+  and so are `aic()`, `bic()` and `mbicv()`, which route through it. The density
+  was accumulated as a running product over up to `d(d-1)/2` edges, so a true
+  log-density below about `-745` underflowed to exactly `0` and the
+  log-likelihood to `-inf`; the accumulator is now in log space. A genuinely
+  zero edge density still gives `-inf`, which is correct. The discrete branch of
+  `scores_full()` differenced two densities floored at `1e-20`, so on such a row
+  both legs were equal and the score came out `0` rather than wrong-and-visible;
+  it now differences the log-densities directly (#770)
 
 * `Bicop::parameters_to_tau()`, `parameters_to_taildep()` and
   `parameters_to_beta()` check the shape of their argument, which used to go
--- a/NEWS.md
+++ b/NEWS.md
@@ -202,6 +202,16 @@ wrong thing.
 
 ### BUG FIXES
 
+* `Vinecop::loglik()` leaves out the observations that have no likelihood
+  instead of returning `NaN` for the whole sample, matching `Bicop::loglik()`;
+  `aic()`, `bic()` and `mbicv()` follow it. Evaluation is unchanged and still
+  propagates `NaN`, which is the house convention: a missing value has no
+  density, but it also has no likelihood to contribute. A log-density of
+  `-inf` is not a missing value and is kept. The per-observation-parameters
+  overload also honors the empty-`u` convention now, reporting the value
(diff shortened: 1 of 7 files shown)


tnagler (commented):
The NEWS text is a bit excessive for such a minimal change, otherwise LGTM.

Issue #769, opened 2026-09-10
ci: add a ThreadSanitizer job (clang, not GCC — GCC 11's libtsan is unusable here)


tvatter:
`VINECOPULIB_SANITIZERS` adds `-fsanitize=address,undefined`, neither of which
detects data races, so the Ubuntu/GNU Debug job would pass on racy code. It did:
the race fixed in #764 — `ThreadPool` reading `stopped_`, `jobs_`, `num_busy_`
and `error_ptr_` without holding its mutex — sat green in CI throughout.

**Use clang.** GCC 11's libtsan mis-tracks the mutex across
`pthread_cond_timedwait`, so any code using `std::condition_variable::wait_for`
loses its happens-before edge and every access that mutex protects is reported
as a race. Reduced to a 20-line, provably correct program (producer writes
`payload`/`ready` under the mutex; consumer reads them under the mutex via
`cv.wait_for`), g++-11 reports one data race and one "double lock of a mutex";
plain `cv.wait` is clean; clang 14 and 16 are clean. A textbook worker pool with
no vinecopulib code gets 9 warnings under g++-11 and 0 under clang-16.

`ThreadPool::wait()` uses `cv_busy_.wait_for`, so a GCC-based job would be all
false positives.

What works, verified on #764's branch:

```
cmake -S . -B build/tsan -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_CXX_COMPILER=clang++-16 -DVINECOPULIB_SANITIZERS=OFF \
  -DCMAKE_CXX_FLAGS="-fsanitize=thread -fno-omit-frame-pointer -g"
cmake --build build/tsan --target test_tools_thread_only
setarch "$(uname -m)" -R build/tsan/bin/test_tools_thread_only
```

- pre-fix header: 3 races (`announce_stop`, and `jobs_.empty()` twice)
- post-fix header: 0 warnings
- full `test_tools_thread` suite: 0 warnings, 9/9 pass

Two practicalities for the workflow: TSan needs its own build directory since it
cannot combine with ASan; and on kernels whose `vm.mmap_rnd_bits` exceeds what
the runtime supports it aborts at startup with `FATAL: ThreadSanitizer:
unexpected memory mapping`, worked around by running under
`setarch "$(uname -m)" -R` (or by lowering the sysctl, which needs root). Worth
keeping the `setarch` wrapper in the job so it does not depend on the runner's
kernel settings.

[...]

Issue #768, opened 2026-09-10
findHeaders.cmake generates 2 of 48 headers when the source path contains a regex metacharacter


tvatter:
The `.ipp` loop in `findHeaders.cmake` was deliberately switched to
`string(REPLACE)` because the source directory is arbitrary and may contain
regex metacharacters (comment at lines 24-25). The `.hpp` loop below it still
uses `string(REGEX REPLACE)` on absolute paths and on `${ipp_file}`
(lines 68-69, 75), so it silently misbehaves for such a path.

With the tree checked out at `.../vine+copulib` and `VINECOPULIB_PRECOMPILED=ON`:

```
configure rc=0                      # succeeds
generated/src/**/*.cpp      33      # the .ipp loop is fine
generated/include/**/*.hpp   2      # should be 48
```

The two survivors are `vinecopulib.hpp` and `version.hpp`, which are copied with
`file(COPY)` rather than through the regex loop.

The failure is confusing rather than obvious. With the generated headers absent,
the compiler falls back to any *installed* copy on the include path — and those
still carry their `#include <...ipp>` line, so each `.ipp` is both inlined and
compiled as its own translation unit:

```
generated/src/bicop/family.cpp:72:1: error: redefinition of
  'std::string vinecopulib::get_family_name(vinecopulib::BicopFamily)'
In file included from /usr/local/include/vinecopulib/bicop/family.hpp:130,
```

So a developer whose checkout path happens to contain `+` or `.` gets a
redefinition error pointing at an unrelated installed header. With no
vinecopulib installed they would instead get "file not found".

Fix: use `string(REPLACE)` in the `.hpp` loop as well.

Found while working on #764.

Issue #767, opened 2026-09-10
ThreadPool stays poisoned after a job throws: later wait() calls rethrow and silently drop queued work


tvatter:
`ThreadPool::error_ptr_` is set by `do_job()` and never cleared, so a pool that
has run one throwing job is unusable for the rest of its life: every subsequent
`wait()` rethrows the same stale exception, and because `wait()` treats a stored
error as a reason to cancel the queue, work pushed afterwards is dropped without
running.

```cpp
ThreadPool pool(4);
pool.push([] { throw std::runtime_error("first job"); });
try { pool.wait(); } catch (const std::exception& e) {}   // "first job" — correct

std::atomic<int> ran{0};
pool.push([&ran]() noexcept { ++ran; });
pool.wait();   // throws "first job" again, and ran == 0
```

Observed on `main` and unchanged by #764.

Not reachable today: the only long-lived pool is `VinecopSelector::pool_`, and a
throwing job there propagates out of `select()` and takes the whole object with
it. It matters for anyone reusing a pool, which the class otherwise supports
(`wait()` exists precisely so a pool can be reused without joining).

The fix is to clear `error_ptr_` when `rethrow_exceptions()` consumes it, but
that needs a decision first: `join()` is `wait()` then `announce_stop()` +
`join_workers()`, so when `wait()` throws today the workers are never stopped.
Whether `join()` should still shut down on the error path is the real question.

PR #774, opened 2026-09-10
Fix the thread-pool error state, the dependency-path and generated-header builds, and add a ThreadSanitizer job

### LINKED ISSUE #766: build: -DEIGEN3_INCLUDE_DIR and -DBoost_INCLUDE_DIRS break the build they are documented to fix
AGENTS.md documents `EIGEN3_INCLUDE_DIR` and `Boost_INCLUDE_DIRS` as the way to
point at a dependency when `find_package` cannot find it. Setting either does
not work: `findDependencies.cmake` skips `find_package`, so the imported target
is never created, but `buildTargets.cmake:7`/`:11` link it unconditionally.

```
$ cmake -S . -B b -DVINECOPULIB_PRECOMPILED=ON -DEIGEN3_INCLUDE_DIR=/usr/include/eigen3
CMake Error at cmake/buildTargets.cmake:7 (target_link_libraries):
  Target "vinecopulib" links to:  Eigen3::Eigen
$ cmake -S . -B b -DVINECOPULIB_PRECOMPILED=ON -DBoost_INCLUDE_DIRS=/usr/local/include
CMake Error at cmake/buildTargets.cmake:7 (target_link_libraries):
  Target "vinecopulib" links to:  Boost::headers
```

Header-only is worse because it is silent: configure and generate both succeed,
and the exported target carries a dangling reference that only breaks consumers.

```
vinecopulibTargets.cmake:65:  INTERFACE_LINK_LIBRARIES "Eigen3::Eigen;wdm;Boost::headers;Threads::Threads"
```

This is the same shape as the `wdm_INCLUDE_DIRS` bug fixed for #755, which
suggests the fix there wants generalizing to Eigen and Boost: define an
interface target from the supplied include directory when the variable is set,
so the link line refers to something either way.

Found while working on #764; nothing in CI covers these variables.

tvatter:
Closes #766, #767, #768, #769, #775. Rebased onto `main` now that #764 merged.

## Fixes

- **#767** — `ThreadPool::error_ptr_` was never cleared, so after one throwing
  job every `wait()` rethrew it and dropped the work pushed since.
  `rethrow_exceptions()` consumes what it reports. That settles the question
  the issue raises about `join()`, which left its workers running when `wait()`
  threw: the wait loop moves into a private `wait_for_jobs()`, and `join()`
  stops and joins before rethrowing.
- **#766** — `-DEIGEN3_INCLUDE_DIR` / `-DBoost_INCLUDE_DIRS` skipped the
  `find_package` that defines the target the link line names. wdm's one-off fix
  (#754) becomes a function used for all three.
- **#768** — the `.hpp` loop matched path-derived patterns as regexes, so a
  path containing `+` or `.` generated 0 of 46 headers. `findHeaders.cmake` uses
  `cmake_path()`, which cannot have that bug rather than having it fixed.
- **#775** — the installed package resolved its dependencies with
  `find_dependency`, so a build given a directory shipped a package that failed
  for consumers. It records those directories and falls back to them.
- **#769** — a ThreadSanitizer job, plus a `packaging` job covering the three
  build-system promises above.

## Also fixed: a data race in `Vinecop::fit`

The TSan job reported one on its first run, pre-existing on `main`. Edges of a
tree are fitted concurrently and each reads `hfunc1`/`hfunc2` columns other
edges write (`class.ipp:718` against `:691`, and `:710` against `:693`).

Across 2800 random structures (d = 3..9, 47600 reads) the read column is always
of higher index than the reading edge, and 25% of reads hit a column another
edge of the same tree writes — so serial order was what made the in-place
scheme correct, and the pool removed that guarantee. With more than one thread
the edges read a snapshot of the previous tree; the serial path is unchanged.

The new test asserts a parallel fit reproduces the serial one, but it did not
fail in fifteen runs of the unfixed code: the read almost always wins. TSan
catches it every time, which is the argument for the job.

## CMake 3.20...4.0

For `cmake_path()`, and the range opts into the policies up to 4.0. Boost
becomes CONFIG-only, since `CMP0167` removes `FindBoost`; the `CMP0074` and
`CMP0144` blocks go with it. Two things the policies broke and this fixes:
`codeCoverage.cmake` used an empty `COMMAND ;`, which `CMP0175` rejects.

## Notes

- The TSan job uses **clang**: GCC's libtsan reports false races on any
  `std::condition_variable::wait_for`, which `ThreadPool::wait()` uses.
- It installs R, because the whole `VinecopTest` fixture skips without Rscript.
  The assertion on the `PASSED` count is what caught that.
- The composite action now exports `BOOST_ROOT` and `Eigen3_ROOT`, removing
  that step from six jobs across three workflows. The
  `EIGEN3_INCLUDE_DIR`/`Boost_INCLUDE_DIR` exports are gone: CMake never read
  them, and the Eigen path in them was wrong.

## Verification

Ubuntu 22.04, GCC 12 and clang 14: 755 tests pass with R parity on; TSan is
clean across the three areas and reports the race when the `fit` change is
reverted; #766 and #768 fail before and pass after; the generated tree is
byte-identical across the `cmake_path()` rewrite; the installed package
resolves its dependencies with `find_package` disabled for all three, and fails
against the previous config.

## Follow-up

#776 — `Vinecop::fit` ignores the `num_threads` its controls carry. Left out:
public API change, four reasonable fixes.

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -116,6 +116,13 @@ runs:
             -DCMAKE_INSTALL_PREFIX="$prefix" \
             -DBUILD_TESTING=OFF
           cmake --build eigen-src/build --config Release --target install
+          echo "Eigen3_ROOT=$prefix" >> $GITHUB_ENV
+      shell: bash
+
+    # find_package reads both through CMP0074, so every caller needs them.
+    - name: Export the dependency roots
+      if: inputs.boost == 'true'
+      run: echo "BOOST_ROOT=${{ steps.boost.outputs.BOOST_ROOT }}" >> $GITHUB_ENV
       shell: bash
 
     - name: Install wdm (pinned to ${{inputs.wdm_ref}})
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -52,7 +52,6 @@ jobs:
       - name: Set environment variables
         run: |
           echo "BOOST_ROOT=${{ steps.install-dependencies.outputs.BOOST_ROOT }}" >> $GITHUB_ENV
-          echo "Boost_INCLUDE_DIR=${{ steps.install-dependencies.outputs.BOOST_ROOT }}/include" >> $GITHUB_ENV
           echo "Eigen3_ROOT=${{ runner.temp }}/eigen" >> $GITHUB_ENV
 
       # Precompiled, so the .ipp files become translation units CodeQL can see.
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -46,14 +46,6 @@ jobs:
           R: false
           lcov: false
 
-      # BOOST_ROOT and Eigen3_ROOT must be exported: find_package picks them
-      # up via CMP0074, ahead of CMake's user package registry, where the
-      # in-workspace eigen-src build tree registers itself.
-      - name: Set environment variables
-        run: |
-          echo "BOOST_ROOT=${{ steps.install-dependencies.outputs.BOOST_ROOT }}" >> $GITHUB_ENV
-          echo "Eigen3_ROOT=${{ runner.temp }}/eigen" >> $GITHUB_ENV
-
       # Precompiled, so the .ipp files become translation units CodeQL can see.
       - name: Build (precompiled, so the .ipp tree is compiled)
         run: |
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -133,8 +133,11 @@ jobs:
             if [ "${{ matrix.cfg.os }}/${{ matrix.cfg.name }}" == "ubuntu-latest/GNU" ]; then
               cmake .. -DVINECOPULIB_STRICT_COMPILER=ON \
                 -DVINECOPULIB_CODE_COVERAGE=ON \
+                -DVINECOPULIB_SANITIZERS=ON \
                 -DCMAKE_BUILD_TYPE=Debug \
                 -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+              ASAN_OPTIONS=detect_leaks=1:detect_stack_use_after_return=1 \
+              UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1 \
               make vinecopulib_coverage
             else
               cmake .. -DCMAKE_BUILD_TYPE=Debug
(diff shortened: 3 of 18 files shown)


tvatter:
Independently reproduced the `Vinecop::fit` race and confirmed this fix, from
the #764 side. My own probe — d=8, structure seed 7, n=400, itau,
`vc.fit(u, ctrl, 8)` — under clang-16 TSan:

| build | result |
| --- | --- |
| `main` | rc=66, **2 data races** |
| this branch | rc=0, **0 warnings** |

**There are two racing pairs, and the description only cites one.** Both are
covered by the fix, so nothing needs changing in the code — but a reviewer
checking whether the fix is complete will want to see that both were found:

```
Write class.ipp:718  (hfunc2.col(edge) = ...)  vs  Read class.ipp:691  (hfunc2.col(m-1))
Write class.ipp:710  (hfunc1.col(edge) = ...)  vs  Read class.ipp:693  (hfunc1.col(m-1))
```

The second is the same bug on the `hfunc1` pair, and it reproduces on a
different structure seed than the one in the description.

The `class.ipp` fix reads correctly to me: snapshot per tree, reads through the
`*_in` references, writes still live, and the serial path binds those references
to the live matrices so it copies nothing. That reproduces serial semantics
exactly — a serial read is either the edge's own column before it writes it, or
a higher-index column still holding the previous tree's value, and the snapshot
supplies both.

~~One non-blocking observation: the threaded path copies all four matrices per
tree even when the data is all-continuous and the two `_sub` matrices are
unused.~~

**Corrected** — that was wrong, and the cost is smaller than I said.
`hfunc1_sub`/`hfunc2_sub` are only assigned under `if (get_n_discrete() > 0)`,
so for all-continuous input they stay default-constructed 0x0 and their two
snapshot assignments are no-ops. Continuous data copies two matrices, not four:
3.2 MB per tree level at d=20/n=10000, into a buffer Eigen reuses after the
first level. Against fitting d(d-1)/2 pair copulas that is noise. Only
mixed-discrete input pays for four, and there the `_sub` matrices were already
allocated. No profiling concern.

A gotcha for anyone repeating the measurement: `fit()` takes `num_threads` as
its third parameter, not from the controls object
(`fit(data, FitControlsBicop, size_t num_threads = 1)`). My first probe set it
on a `FitControlsVinecop` and passed that as `controls`, so it ran serially and
reported 0 races on `main` — which looked like a refutation until I noticed the
probe was wrong.

PR #777, opened 2026-09-14
feat(bicop)!: view a fitted copula under different variable types

tvatter:
Stacked on #771.

## Why

`Bicop::as_continuous()` returned the same copula with neither variable declared
discrete. There was no way to go the other direction without `set_var_types()`,
which mutates an object the caller may not own.

That gap has a downstream cost. pyvinecopulib's `DiscreteBicop` builds the
mixed-discrete surface for a pair by differencing four `cdf` values — the route
`AbstractBicop` took until #771. Since #771 a kernel copula reads each atom's
probability off the interpolation grid instead, so the two disagree by ~1e-9 at
the atom widths a vine's inner trees reach, and a wrapper holding a compiled
`Bicop` has no way to ask for the better route.

## What changed

Per @tnagler's suggestion, one method rather than a pair:

```cpp
Bicop with_var_types(const std::vector<std::string>& var_types = { "c", "c" }) const;
```

It replaces `as_continuous()` and goes in either direction. The model is
untouched — only the declaration the evaluation methods read changes — so a
fitted copula can be evaluated on a continuous, discrete or mixed edge without
being refitted.

**Breaking, and the old spelling is the default argument:** `bc.as_continuous()`
becomes `bc.with_var_types()`. NEWS and the migration guide both record it.

`BicopView::as_continuous()` keeps its name — it is a non-owning view with no
types to supply, and it is internal.

`set_var_types()` already validates, so an ill-formed declaration throws from
there. `as_continuous()` was also undocumented, which downstream meant an empty
docstring, since these comments are lifted verbatim into the Python API text.

## Test

`discrete.with_var_types_round_trips_the_variable_types` pins that the view and
a mutated copy are bit-identical on `pdf` / `hfunc1` / `hfunc2`, that the source
is left alone, that the two directions are inverse, and that a bad declaration
throws.

Full suite: 765 tests, all passing. clang-format 14.0.6 and codespell 2.4.3 clean.

No NEWS entry beyond the breaking-change line — this rides with #771.


tnagler:
maybe `as_continuous()` and `as_discrete()` are not the right names if this is used generally. maybe just a single method called `with_var_types()` or so?

PR #792, opened 2026-09-21
fix(stats): keep pnorm finite-valued at infinite arguments under Eigen 5

### LINKED ISSUE #784: Segfault with Eigen 5+
When updating the conda recipe for 1.0.0, CI initially failed: https://github.com/conda-forge/pyvinecopulib-feedstock/actions/runs/35360781908/job/105651226083

The following fixed it:

```YAML
host:
  ...
  - eigen >=3.4,<5
```

We need to investigate this.

tnagler:
Fixes #784.

## What was wrong

Eigen 5's `erf` for `double` (`generic_fast_erf<double>` in
`SpecialFunctionsImpl.h`) has no input clamp, unlike its `erfc`: an infinite
argument goes through `inf * inf` and `inf - inf` and comes out NaN, where
Eigen 3.4 returned ±1. This is independent of vectorization (same result
with `EIGEN_DONT_VECTORIZE`) and also affects finite arguments around 1e300.

`tools_stats::pnorm` is `0.5 * (1 + erf(x / sqrt 2))`. `find_latent_sample`
evaluates `pnorm(qnorm(1) + b) = pnorm(inf)` for the top level of every
discrete variable, so the upper box bounds became NaN. The NaN-to-`size_t`
cast in `BoxCovering::get_box_indices` then produced a garbage cell index
and the covering was read past its end. On Linux with ASan this is a
heap-buffer-overflow in `test_tools_stats.find_latent_sample_is_deterministic`;
on Windows it is the access violation reported in #784 (the pyvinecopulib
tests exercise discrete models).

The Linux job failure in #784 is a different problem: pyvinecopulib's
libclang docstring extractor cannot parse the GCC 15 intrinsics headers that
Eigen 5 pulls in. That is a downstream tooling issue (e.g. pass
`-DEIGEN_DONT_VECTORIZE` to the mkdoc parse), not a library bug.

## The fix

- `pnorm` clamps its argument to [-40, 40] before calling `erf`. The CDF is
  exactly 0 or 1 in double precision beyond that range, so results are
  bit-identical on Eigen 3.4 (checked on a 2001-point grid against
  `0.5 * erfc(-x / sqrt 2)`, max difference 1.1e-16) and finite on Eigen 5.
  NaN passes through unchanged, as before.
- `BoxCovering` maps coordinates through a clamped `cell` helper, so a
  coordinate of exactly 1 (which `pnorm` can produce far in the tail) lands in
  the last cell instead of one past it, and bounds outside the unit square
  are capped at the boundary cells. This is the same defect class and would
  otherwise remain a latent out-of-bounds write in `swap_sample`.

## Verification

- `main` built against Eigen 5.0.1 with `-DVINECOPULIB_SANITIZERS=ON`
  (Debug, GCC 12): before the fix, ASan heap-buffer-overflow as described;
  after the fix, the full suite passes (769 tests, 106 R-parity skips) with
  no sanitizer reports.
- New tests: `pnorm_saturates_in_the_tails` (±inf, ±1e300, ±50, NaN) and
  `box_covering_handles_the_boundary`.

I could not confirm from the Eigen commit history whether upstream has fixed
`erf` at infinity since 5.0.1; the clamp makes the library independent of
that either way.


tvatter:
LGTM but I'd like to verify it actually fixes the issue I noticed in vinecopulib before claiming it actually fixes #784. Did you?

tvatter:
Verified in https://github.com/vinecopulib/pyvinecopulib/pull/352

